PR #8, opened 2015-02-27
set zero to be imaginary (for old assumptions)

skirpichev:
see https://github.com/sympy/sympy/pull/8745

- [x] rebase
- [x] workaround test_issue_4149


skirpichev:
(this is reply on [this post](https://github.com/sympy/sympy/issues/7649#issuecomment-195954735))

 > @skirpichev are you still working on this? I would like to work on this if you're not

@AnishShah, I'm not working on this, as I already did all my work on this and that was merged.  Feel free to work on this for sympy and/or reuse my work.  But keep in mind that my code is not in public domain, please respect license terms.

Sorry for late answers, but I don't get notifications from sympy repos even if you @mention me.

PR #23, opened 2015-03-08
Removed physics module

skirpichev:


aktech:
BTW, there is some dependency on physics module, for example: [this](https://github.com/skirpichev/omg/blob/master/sympy/solvers/solvers.py#L168)

skirpichev:
Yes, but first I look on tests timing.  It seems, I still should split tests.

PR #39, opened 2015-03-26
Use srepr instead of sstr for __repr__ printing

skirpichev:
see https://github.com/skirpichev/omg/issues/32

- [x] fix tests
- [x] document repr behavior in docs
- [x] fix polys repr methods (e.g. SubModule.__repr__ returns things like '<x>')
- [x] rebase

most test failures because container’s \__str__ uses contained objects’ \__repr__.  I.e. see [this](http://stackoverflow.com/questions/1436703/difference-between-str-and-repr-in-python/2626364#2626364).

See also:
https://www.python.org/dev/peps/pep-3140/
https://groups.google.com/forum/#!topic/python-ideas/ChVbmWhlqJA

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -43,3 +43,16 @@ def check_disabled(request):
         if (V(pytest.__version__) < '2.6.3' and
                 pytest.config.getvalue('-s') != 'no'):
             pytest.skip("run py.test with -s or upgrade to newer version.")
+
+
+@pytest.fixture(autouse=True, scope='session')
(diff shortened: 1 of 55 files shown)


skirpichev:
most test failures because container’s __str__ uses contained objects’ __repr__.

skirpichev:
https://www.python.org/dev/peps/pep-3140/
https://groups.google.com/forum/#!topic/python-ideas/ChVbmWhlqJA

PR #122, opened 2015-08-11
Remove new assumptions

skirpichev:


skirpichev:
Perhaps, someday we can reintroduce it, but now the presence of new assumptions will just confuse users, see e.g. [this](https://groups.google.com/forum/?fromgroups=#!topic/sympy/AJVZdWnYQww).

PR #139, opened 2015-12-01
Catch NotImplementedError from gruntz

skirpichev:


Upabjojr:
@skirpichev Are you sure that you've solved https://github.com/sympy/sympy/issues/8061 ?

Running `limit(4**(acos(1/(1+x**2))**2)/log(1+x, 4), x, 0)` I get `oo`, while Mathematica gives `0`.

Upabjojr:
Sorry, wrong argument order in `log`. SymPy's equivalent of `log(x, y)` in Mathematica is `Log[y, x]`, not `Log[x, y]`.

skirpichev:
> Sorry, wrong argument order

So, you don't question the limit value, @Upabjojr?

But there is a bug with log (unevaluated two-arg version) printing for Mathematica.

Upabjojr:
> So, you don't question the limit value, @Upabjojr?

Well, I was just cross-checking the limit value with Mathematica. I guess issue 8061 has `1+x` and `4` swapped in `log`.

> But there is a bug with log (unevaluated two-arg version) printing for Mathematica.

Yes, but it will raise an exception if you use `evaluated=False` in `log`. So not a big deal.

skirpichev:
> I guess issue 8061 has 1+x and 4 swapped in log.

Perhaps.  I don't see anything wrong on my side.   Regression test match provided in the bugreport input.

And Mathematica's output is same as diofant's:
```
In[1]:= Limit[4^(ArcCos[(x^2 + 1)^(-1)]^2)*Log[4]/Log[x + 1], x -> 0, 
 Direction -> -1]

Out[1]= \[Infinity]
```
Input expression here was generated literally from bugreport with ``mathematica_code()``.

> it will raise an exception if you use evaluated=False in log

Indeed.

PR #191, opened 2016-01-14
Add quick tests to checksol: is_nonzero

skirpichev:


skirpichev:
@smichr, I think the test for `solve(sqrt(x + root(x, 3)) + root(x - y, 5), y)`, see 312c1af, was broken in the sympy as well...

smichr:
Indeed. I have that fix in one of the branches that I am working on but am only now getting back to it after working on some other issues that have come up. Stymied on what to do about theano not wanting certain re-writings of And args at the moment.

smichr:
No, that was a different test that I was working on. Although you have changed the test to another equation that does work, the original equation is still not being solved correctly so there is a hidden problem somewhere.

skirpichev:
On Sat, Jan 16, 2016 at 07:12:41AM -0800, Christopher Smith wrote:
>    No, that was a different test that I was working on. Although you have
>    changed the test to another equation that does work, the original equation
>    is still not being solved correctly so there is a hidden problem
>    somewhere.

I mean, your test _answer_ was wrong.

smichr:
Yes, the answer is wrong, but in your modifications you change the answer to correspond to a *new equation*. This has not solved the underlying issue of obtaining a solution (or rejecting a solution) of the original equation.

skirpichev:
On Sat, Jan 16, 2016 at 08:14:45AM -0800, Christopher Smith wrote:
>    Yes, the answer is wrong, but in your modifications you change the answer
>    to correspond to a new equation. This has not solved the underlying issue
>    of obtaining a solution (or rejecting a solution) of the original equation.

Good point, perhaps xfailed test should be added too (solve != []).  Not
sure if it worth keeping also the modified test (as I did), but now it's
correct at least.

smichr:
I was wrong about saying that the solution to the new equation is right. Whether the sign is positive or negative in the original expression, the solutions are not uniformly 0 and equals will verify that:

```python
>>> eq
sign*(x - y)**(1/5) + sqrt(x**(1/3) + x)
>>> sol = {}
>>> for s in (-1, 1):
...   expr = eq.subs(sign, s)
...   ans = solve(expr, y)
...   z = expr.subs(y, ans[0])
...   sol[s] = z
...   print s, z.equals(0)
...
-1 False
1 False
>>> sol[-1].subs(x,1)  # it's 0 sometimes
0
>>> sol[-1].subs(x,-1).n(2)  # but not always
-0.48 + 1.1*I

>>> sol[+1].subs(x,0)  # it's got a trivial zero
0
>>> sol[+1].subs(x,1).n(2)  # but that's all
2.6 + 0.83*I
>>> sol[+1].subs(x,-1).n(2)
1.4 + 1.3*I
```

smichr:
An additional problem is that in one case, the solution passes by an explicit warning request:

```python
>>> solve(eq.subs(sign,1), y, warn=True)
sympy\solvers\solvers.py:309: UserWarning:
        Warning: could not verify solution {y: x + (x**(1/3) + x)**(5/2)}.
  warnings.warn("\n\tWarning: could not verify solution %s." % sol)
[x + (x**(1/3) + x)**(5/2)]
>>> solve(eq.subs(sign,-1), y, warn=True)
[x - (x**(1/3) + x)**(5/2)]
```

skirpichev:
I fact, I get sometimes False and sometimes None.

Sympy (non-fork) session, for example:
```
In [1]: eq2 = sqrt(x**(1/3) + x) - (x - y)**(1/5); eq2
Out[1]: 
   ___________            
  ╱ 3 ___        5 _______
╲╱  ╲╱ x  + x  - ╲╱ x - y 

In [2]: sol2 = solve(eq2, y); sol2
Out[2]: 
⎡               5/2⎤
⎢    ⎛3 ___    ⎞   ⎥
⎣x - ⎝╲╱ x  + x⎠   ⎦

In [3]: sb = eq2.subs(y, sol2[0])

In [4]: sb
Out[4]: 
                     ________________
   ___________      ╱            5/2 
  ╱ 3 ___        5 ╱  ⎛3 ___    ⎞    
╲╱  ╲╱ x  + x  - ╲╱   ⎝╲╱ x  + x⎠    

In [5]: sb.equals(0)

In [6]: sb.equals(0)
Out[6]: False
```

Same in my fork:
```
$ ipython --profile omg

In [1]: eq2 = sqrt(x**(1/3) + x) - (x - y)**(1/5)

In [2]: sol2 = solve(eq2, y)[0]

In [3]: s2 = eq2.subs(y, sol2); s2
Out[3]: 
                     ________________
   ___________      ╱            5/2 
  ╱ 3 ___        5 ╱  ⎛3 ___    ⎞    
╲╱  ╲╱ x  + x  - ╲╱   ⎝╲╱ x  + x⎠    

In [4]: s2.equals(0)
Out[4]: False

In [5]: s2.equals(0)

In [6]: s2.equals(0)
Out[6]: False
```

Numerical evaluation in random rational/integer points suggests that the answer might be True, e.g.:
```
In [11]: s2.subs(x, 1)
Out[11]: 0

In [12]: s2.subs(x, 1/2)
Out[12]: 0
```

skirpichev:
```
>>> sol[-1].subs(x,-1).n(2)  # but not always
-0.48 + 1.1*I
```

Oh, indeed, I miss that example.  So, it's non-0.

skirpichev:
As I can see from https://github.com/sympy/sympy/issues/10398
- you think, there is no problems...  Ok, understood.

Not new for me, unfortunately.

On Sat, Jan 16, 2016 at 8:11 PM, Christopher Smith <notifications@github.com

PR #217, opened 2016-02-05
Use codecov instead of coveralls

skirpichev:


skirpichev:
Nice tool, @asmeurer thank you for hint in sympy/sympy#10519.
But it counts coverage somehow different, c.f.:
https://codecov.io/github/skirpichev/omg?ref=master
and (with same .coveragerc):
https://coveralls.io/github/skirpichev/omg
Maybe the first codecov submission (short gmpy run) had been ignored,
but this shouldn't add too much difference (gmpy test add only few unique lines to the whole coverage).

gxyd:
Do you need to run the codcov manually or is it run automatically? Also does it lead to doubling the run time of test suite? Is this run on all versions of Python?

skirpichev:
> Do you need to run the codcov manually or is it run automatically?

You could take look on any recent PR, instead of asking dumb questions.

> Also does it lead to doubling the run time of test suite?

Not to doubling, perhaps, but it noticeably increase running time.  That's why I don't gather coverage stats for slow and xfail'ed tests.  See .travis.yml.

> Is this run on all versions of Python?

There shouldn't be version-specific workarounds.  So, no.  Again, see .travis.yml.

gxyd:
> You could take look on any recent PR, instead of asking dumb questions.

I looked at them. I saw codecov was run on all (the ones I saw) PR's. It still could have been the case that you were manually starting them all.

skirpichev:
> It still could have been the case that you were manually starting them all.

Any thoughts why you would like to manually trigger coverage build and submit results for a **required** check?  That's silly.

gxyd:
Not unless every pull request is a doctest change.

PR #246, opened 2016-03-11
Use limit in hyperexpand

skirpichev:
This alone could solve summation problem in skirpichev/omg#236, without using (not a bad idea anyway) gosper_sum first.


skirpichev:
@seblabbe, you may be interested in this patch, as the problem was reported in
https://groups.google.com/forum/?fromgroups=#!topic/sympy/ZI_D-eG1UfE

Sorry for off-list answer, but I can't post this solution to sympy list.

seblabbe:
Great, thanks for the fix!

PR #359, opened 2016-10-16
Beta 3

skirpichev:


diofant/core/tests/test_cache.py
```diff
@@ -96,3 +98,20 @@ def test_nocache(clear_imports, monkeypatch):
     y = Symbol('y')
     result = s.subs(sqrt(x**2), y)
     assert result == 1/y
+
+    # problem from https://groups.google.com/forum/#!topic/sympy/LkTMQKC_BOw
+    # see commit c459d18
+    a = Symbol('a', positive=True)
+    f = exp(x*(-a - 1))
+    g = sinh(x)
+    r = f*g  # not raises
```
skirpichev: @Ruediger-Braun, you may be interested in fix for the problem, mentioned in https://groups.google.com/forum/#!topic/sympy/LkTMQKC_BOw

This issue was fixed in Diofant 0.8.  By the way, it seems to be fixed in the recent SymPy as well.

Ruediger-Braun: @skirpichev Both problems are indeed not present in diofant.  They are still present in sympy 1.0.1_dev

Thank you very much.

skirpichev: > They are still present in sympy 1.0.1_dev

Hmm, indeed.  Sorry for this misinformation.

This problem is hash-related (thus, sometimes sympy is correct and not raises an exception). Perhaps I have not run tests in sympy enough times.

skirpichev:
@Shikhar1998, regarding [your comment](https://github.com/sympy/sympy/issues/2865#issuecomment-285355321).

> 1. can you please clarify what's the problem above

The problem is a wrong answer (or answers, in fact).

> 2. what output is expected

As I said, an error, perhaps.  This is for SymPy, I see no options.  In the Diofant - unevaluated limits returned, this is more logical.

> As O(x) can be written as O(1)*x so applying limit x->0 will give an output zero.

Yes, but this example assume using Big-O notation around x=0.  Problem limits use other point, namely x=oo.  Please learn what the Big-O symbol is.


Sorry for late answers, but I don't get notifications from sympy repos even if you \@mention me.

Issue #376, opened 2016-11-11
problem with derivative and chain rule


RiccardoRossi:
Dear Developers,

i have been trying out Diofant on my computer (ubuntu 16.04) and i have a little issue with chain rule in derivative.

The issue is, perhaps not surprisingly, the same on sympy and diofant, still i thought it might be worth to post it here

from diofant import *

u,B = symbols('u B')
e = B*u #here i use an expression instead of a function

s = Function('s')(e)
print(diff(s,u).doit())

gives
B*Subs(Derivative(s(_xi_1), _xi_1), (_xi_1,), (B*u,))

while i would expect smthg like 
B*Derivative(s(e),e)

please be patient if the question/issue is wrong.

full question in the sympy forum is here:
https://groups.google.com/forum/#!msg/sympy/ogz52F5Ol5c/x2Mbwq41BAAJ

skirpichev:
> while i would expect smthg like
> B*Derivative(s(e),e)

Actually, what I get here - is exactly that you expect.  Maybe with pretty-printing in usual mathematical notation it would be clear:
```
In [6]: diff(s,u).doit()
Out[6]: 
  ⎛ d        ⎞│      
B⋅⎜───(s(ξ₁))⎟│      
  ⎝dξ₁       ⎠│ξ₁=B⋅u
```

Please keep in mind, that in the diofant - there is no derivative operator, like Derivative in Mathematica.  You have to provide arguments, with respect to which you differentiate your expression.  That's why we have a Subs (unevaluated substitution) instance here and then - a Derivative instance inside, with a dummy variable.

> please be patient if the question/issue is wrong.

Don't worry.  Please take look on docstrings of classes Derivative, Subs; diff and doit methods.

Let me know if think that the answer is not clear enough.

> full question in the sympy forum is here:
> https://groups.google.com/forum/#!msg/sympy/ogz52F5Ol5c/x2Mbwq41BAAJ

It's not entirely clear for me what you are trying here to implement.

RiccardoRossi:
Dear Sergey,
    thank you very much for your attention. I managed to reproduce the output u posted using jupyter-notebook, unfortunately it does not fit to what i am trying to achieve.

The thing is that i am developing a small tool to obtain the c++ implementation of a "Finite Element" starting from the symbolic definition of the underlyining functional.
From the point of view of symbolic calculus i think this is a rather simple thing since it only requires taking derivatives of polynomial functions.

[...]

PR #378, opened 2016-11-12
Simple fix for substitutions of Derivatives

skirpichev:


skirpichev:
@RiccardoRossi, does it work for you as a workarround in #376?

RiccardoRossi:
Well,
    It was born as a workaround, however
if this work I would now consider it as a first choice solution since I
wouldn t have to do a subs "a posteriori".

However it still throws when e=B*u ... And to be honest I don't have a clue
on how to cure that.

Regards, and thx again for your time!

Riccardo

skirpichev:
> if this work I would now consider it as a first choice solution since I wouldn t have to do a subs "a posteriori".

Please test that this solution is working for your application, if you can.

PR #409, opened 2017-01-27
Symbolic Matrix rank

skirpichev:
- [ ] avoid workarounds.  Probably, this will require implementing \_min/_max state variables to indicate lower/upper bounds for the rank.  Use FiniteSet's and Range?


skirpichev:
reopened in https://github.com/diofant/diofant/pull/878

PR #419, opened 2017-02-02
Filter out zeros in GroebnerBasis constructor

skirpichev:
- [x] check that solution is full (btw, Mathematica's Solve is too slow to this)


skirpichev:
@vog, just to resolve your confusion [here](https://github.com/sympy/sympy/issues/12114#issuecomment-277672271).

Your issue was fixed in the [Diofant](https://github.com/diofant/diofant/) project, **not in** the SymPy.  You can follow to [commit page](https://github.com/skirpichev/diofant/commit/c51c9e43c729c78c402611fd66fe0a6bdaae3f1f) and it shows you this PR number #419.  Apparently, this PR **is not** against SymPy repository.

PR #442, opened 2017-02-20
rsolve_hyper: take into account degenerate solutions

skirpichev:
@klemvor, @pedritomelenas, let me know if you have some comments.

### DIFF
--- a/diofant/solvers/recurr.py
+++ b/diofant/solvers/recurr.py
@@ -621,11 +621,12 @@ def rsolve_hyper(coeffs, f, n, **hints):
             if z.is_zero:
                 continue
 
-            (C, s) = rsolve_poly([ polys[i]*z**i for i in range(r + 1) ], 0, n, symbols=True)
+            sol, syms = rsolve_poly([polys[i]*z**i for i in range(r + 1)],
+                                     0, n, symbols=True)
+            sol = sol.collect(syms)
+            sol = [sol.coeff(_) for _ in syms]
 
-            if C is not None and C is not S.Zero:
-                symbols |= set(s)
-
+            for C in sol:
                 ratio = z * A * C.subs(n, n + 1) / B / C
                 ratio = simplify(ratio)
 
--- a/diofant/solvers/tests/test_recurr.py
+++ b/diofant/solvers/tests/test_recurr.py
@@ -207,8 +207,8 @@ def test_rsolve_raises():
 
 def test_sympyissue_6844():
     f = y(n + 2) - y(n + 1) + y(n)/4
-    assert rsolve(f, y(n)) == 2**(-n)*(C0 + C1*n)
-    assert rsolve(f, y(n), {y(0): 0, y(1): 1}) == 2*2**(-n)*n
+    assert rsolve(f, y(n)) == 2**(-n + 1)*C1*n + 2**(-n)*C0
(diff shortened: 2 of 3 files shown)


pedritomelenas:
Thanks for the fix.
```
In [1]: from diofant import *

In [2]: a=Function('a')

In [3]: n=Symbol("n", integer = True)

In [4]: rsolve(a(n)-2*a(n-1)-n,a(n))
Out[4]: 
 n           
2 ⋅C₀ - n - 2

In [5]: rsolve(a(n)-2*a(n-1)-n,a(n),[1])
Out[5]: 
   n        
3⋅2  - n - 2
```
Which seems to be correct. 
I have tried one that it is derived from the above equation:
```
rsolve(a(n) - 2*a(n - 3) + 5*a(n - 2) - 4*a(n - 1),a(n),[1,3,8])
```
And get no answer, while it has a solution (am I doing something wrong here?). Please, have a look at the end of this [notebook](https://github.com/pedritomelenas/LMD/blob/master/Naturales/Naturales.ipynb) (it is in Spanish, but the equations are easy to follow; please let me know if you need a translation).

Thank you very much for caring about this.
Pedro

skirpichev:
> And get no answer, while it has a solution

Hmm, strange.  That's what I get on this branch:
```
In [5]: rsolve(a(n) - 2*a(n - 3) + 5*a(n - 2) - 4*a(n - 1),a(n),[1,3,8])
Out[5]: 
   n        
3⋅2  - n - 2
```
Actually, this test was added to my test suite in this PR (the only change is list->dict in initial conditions).  So, it should work.

Be sure you are testing this branch, not the master branch of the diofant repo.  This pr is not merged yet.

skirpichev:
> Please, have a look at the end of this notebook

It looks, there are other mentioned here examples, which fail on SymPy, e.g.:
```python
In [45]: rsolve(a(n+2)-5*a(n+1)+6*a(n)-n,a(n))
Out[45]:
2**n*C0 + 3**n*C1 + C0*n
```

@pedritomelenas, I will double check tomorrow, but right now I don't see any failing examples from that notebook.  Above example also returns now a correct answer:
```python
In [1]: a=Function('a')

In [2]: rsolve(a(n+2)-5*a(n+1)+6*a(n)-n,a(n))
Out[2]: 
 n       n      n   3
2 ⋅C₀ + 3 ⋅C₁ + ─ + ─
                2   4

In [3]: (a(n+2)-5*a(n+1)+6*a(n)-n).subs(a, Lambda(n, _2))
Out[3]: 
   n         n + 1       n + 2         n         n + 1       n + 2   
6⋅2 ⋅C₀ - 5⋅2     ⋅C₀ + 2     ⋅C₀ + 6⋅3 ⋅C₁ - 5⋅3     ⋅C₁ + 3     ⋅C₁

In [4]: simplify(_)
Out[4]: 0
```

pedritomelenas:
Yes, you are right. I was not using this branch. 
I will try this branch later.

Thanks.

pedritomelenas:
Works! Thanks!

skirpichev:
I'll add last example as a regression test (btw, this one already fixed in the master branch).

Let me know if I miss something else from the notebook.

pedritomelenas:
I will run more tests. Thanks!

Issue #445, opened 2017-02-21
Clarify the license of Diofant


certik:
The BSD license in SymPy:

https://github.com/sympy/sympy/blob/d29d13a556e9e2c5622edfd345e4e99b9089a980/LICENSE

requires to "... retain the above copyright notice, this list of conditions ...", however the Diofant license modified the `c` clause:

https://github.com/diofant/diofant/blob/e97a1ee507cc66e147d92160c5c24a5cb88eb54b/LICENSE

I think that might be problematic, but I am not a lawyer. Regarding adding things to the copyright, I think that's probably fine.

@skirpichev, it would be helpful if you could please clarify the situation. The standard way of dealing with this, and also the way I dealt with it in https://github.com/sympy/sympy/pull/10531, is to simply copy the original license, intact, somewhere, and then add a new license, e.g. in SymPy, I copied your license, intact, and then kept the SymPy license, and finally I clarified which patches are covered by your license and which are not. So one way that would make this clear is if you copied the SymPy license, unmodified, and then added your own license, and further indicate which code or which patches are covered by each. Or if it's too complicated, I think you can just say "this project uses code from SymPy that was covered by SymPy's license:" and paste the license there or something like that.

That way there is no confusion to what code you own the copyright to, and what you don't. You own the copyright to your own code in Diofant, but you do not own copyright to SymPy's code (except your own patches that are merged in SymPy).

skirpichev:
> Regarding adding things to the copyright, I think that's probably fine.

I believe so.  The difference in c) clause is emphasized here: "Neither the name **of Diofant**, or SymPy nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission."

In fact, I can drop bold'ed part, if you really insist.

> The standard way of dealing with this

Standardized by whom?

I only follow license terms, which don't changed.  (For just one commit - the copyright statement of SymPy Development Team was changed to include 2017 year...)  And additional copyright statement was added for this derived product.

> is to simply copy the original license, intact, somewhere, and then add a new license

And than you copy this license intact, append to your license.  Next round I will take your new license, append to mine and so on...

> further indicate which code or which patches are covered by each

[...]

PR #307, opened 2016-07-05
Support IVP for dsolve

skirpichev:
based on https://github.com/sympy/sympy/pull/11264

- [x] coverage tests
- [x] partial solutions
- [x] ~~better place for solve_ics? (i.e. not \_helper_simplify)~~
- [x] rebase
- [x] ask for permission to merge
- [x] rename kwarg to be consistent with rsolve?
- [x] adapt existing test (e.g. Wester's) to use initial conditions

PR #446, opened 2017-02-22
Correct check in try_polynomial()

skirpichev:
@jksuom, let me know if I can't merge PR's from your SymPy fork on terms of the SymPy license.  I.e. parts of your non-merged-yet work.  I got permission for [Aaron's stuff](https://github.com/diofant/diofant/pull/307#issuecomment-247626007) so far.

- [x] close https://github.com/diofant/diofant/issues/445


jksuom:
> let me know if I can't merge PR's from your SymPy fork

You can merge. I have no objections.

(BTW, in your method of dealing with shift operators, it seems to me that `a == b` may occur. since common parameters are not removed, but that should not affect the final result.)

skirpichev:
Thank you for permission and review, I'll check code flow again to be sure.

skirpichev:
> it seems to me that a == b may occur.

@jksuom, why do you think so?  ``ReduceOrder`` (which is called by ``_reduce_order()`` helper) should work for ``n=ai-bj=0``.  Your solution sympy/sympy#12182 seems to be correct to, but I have some objections against your first commit.
1) it seems to be a partial duplicate of ``_reduce_order``
2) some reduction operators will be missed in your case, because you preprocess input for ``_reduce_order``

On another hand, difference with mpmath's defaults, [mentioned here](https://github.com/sympy/sympy/pull/10531#issuecomment-281806140) - does matter.  One possible solution: add new evalf helper for ``hyper()``, which will set ``eliminate`` kwarg.  Or even change mpmath's defaults.

Also, new results for cases like ``hyper((-1, 1), (-2,), z)`` are consistent with Mathematica's output.  So, probably it's equally correct, unless you think that the result will be inconsistent for some parameters.

cc @certik

jksuom:
The comment about `a == b` was there only to remind that common parameters had been eliminated. Otherwise the inequality `a > b` should have been written in the form `a >= b`.

> it seems to be a partial duplicate of _reduce_order

Yes, it is. I was thinking of simplifying _reduce_order later in another commit. Using preprocessed parameters (and omitting some parts) to avoid repetition.

> some reduction operators will be missed in your case,

I think that it would only happen in case no shift operators are wanted.

I was thinking of adding kwarg `eliminate` to hyperexpand.

skirpichev:
Looking further, mpmath behaviour wrt "eliminate" flag looks odd.  No matter which definition is used for common nonpositive integer upper and lower parameters - result should be same with or without expansion.

skirpichev:
Ok, at least Fredrik - positive for changing defaults.  This could be merged meanwhile, as a temporary workarround.

PR #526, opened 2017-08-06
New pattern matching with matchpy

skirpichev:
See Also
-----------

https://github.com/sympy/sympy/wiki/Proposal-for-a-new-pattern-matching
sympy/sympy#12978


skirpichev:
@Upabjojr, I've seen that you are [trying to include](https://github.com/sympy/sympy/issues/20743) the matchpy as required dependency of SymPy,  Correct me, if I'm wrong, but it seems your matchpy connector is not capable enough to provide all Basic.match() functionality.

For example, I don't see how something like this, `(2*x).match(a*x)={a: 2}` may work right now.  I think, to mark atomic expressions - the matchpy.Atom API should be changed like matchpy.Operator.  So, for instance, we could do something like `matchpy.Atom.register(AtomicExpr)`.

skirpichev:
UPD.  Oh, I've found the source of trouble.  The `Mul` is not commutative per default, so it's important that arguments in the expression and in the pattern are ordered consistently.  Thus, I think the matchpy doesn't fit well in the current model for commutative/noncommutative expressions of the SymPy/Diofant.  There should be NonCommutativeMul class to match the current matchpy API.

Also, I'm not sure how to fix matching for identities.  E.g. `(2*x).match(p*x + q)`, where for `q` we should have `0`.

Upabjojr:
> For example, I don't see how something like this, `(2*x).match(a*x)={a: 2}` may work right now.

```python
In [1]: from matchpy import *

In [2]: from sympy import *

In [3]: from sympy.utilities.matchpy_connector import *

In [4]: a = WildDot("a")

In [5]: ;var x
Out[5]: x

In [6]: matcher = ManyToOneMatcher()

In [8]: matcher.add(Pattern(a*x), "a times x")

In [10]: next(iter(matcher.match(2*x)))
Out[10]: ('a times x', {'a': 2})
```

> The `Mul` is not commutative per default, so it's important that arguments in the expression and in the pattern are ordered consistently.

For now, the `matchpy_connector` objects assume it's commutative. 

> There should be NonCommutativeMul class to match the current matchpy API.

There is currently an issue about this: https://github.com/sympy/sympy/issues/19992

> Also, I'm not sure how to fix matching for identities. E.g. `(2*x).match(p*x + q)`, where for `q` we should have `0`.

MatchPy supports `optional=...` in order to define _optional_ values in case they are not matched:

```python
In [18]: p = WildDot("p", optional=S(1))

In [19]: q = WildDot("q", optional=S(0))

In [20]: matcher = ManyToOneMatcher()

In [21]: matcher.add(Pattern(p*x + q), "matching with default values")

In [22]: next(iter(matcher.match(x + 3)))
Out[22]: ('matching with default values', {'p': 1, 'q': 3})

In [23]: next(iter(matcher.match(2*x)))
Out[23]: ('matching with default values', {'p': 2, 'q': 0})
```

It would be nicer if you could just tell that the optional value is the identity element of the current node, as Wolfram Mathematica does (instead of specifying it manually).

Upabjojr:
> Thus, I think the matchpy doesn't fit well in the current model for commutative/noncommutative expressions of the SymPy/Diofant.

Maybe [this algorithm](https://github.com/HPAC/matchpy/blob/d542872f804e22590bfbee73f3121f80a99878a2/matchpy/matching/bipartite.py#L269) could be updated to keep the order of non-commutative arguments? It's the iterator for the commutative matcher.Though I'm not sure how much work would be needed to modify MatchPy, as that algorithm is quite complex, plus the structure should be adapted to remember the permissible permutations.

skirpichev:
> matcher.add(Pattern(a*x), "a times x")

Oh, my bad.  Probably, I used `p = Wild('p')`, instead of `a` to get wrong order of arguments in the pattern.  And indeed, for a commutative case everything is working as expected, regardless of wildcards naming.

> MatchPy supports optional=... in order to define optional values in case they are not matched

I saw this, but it doesn't look as a good workaround.  For Mul and Add you, probably, will want different optional= values.

> It would be nicer if you could just tell that the optional value is the identity element of the current node

Yes, I think so.  Probably, this can be part of the Operator class API, as now we have `create_operation_expression()`, `op_len()` and so on.  No-args function to return the default?

Upabjojr:
> > It would be nicer if you could just tell that the optional value is the identity element of the current node
> 
> Yes, I think so. Probably, this can be part of the Operator class API, as now we have `create_operation_expression()`, `op_len()` and so on. No-args function to return the default?

https://github.com/HPAC/matchpy/issues/69

skirpichev:
@Upabjojr, beware - I don't get notifications from the SymPy repo.  Sometimes, I look on SymPy's PR's, but not so often and usually restrict my attention for major directions of the current Diofant's development.  Right now, it's polys/domains modules.  (Unfortunately, I don't think that integrating matchpy for pattern matching is a viable target in the near future.  Major obstacle is the difference in handling commutativity.  Not sure how to solve it.  Perhaps, the Diofant should adapt the MatchPy model to implement operators.)

So, if you want to ping me for some feedback, please do this, for example, in the relevant Diofant's PR/issue, instead of [doing this](https://github.com/sympy/sympy/pull/20794#issuecomment-760153302).

Regarding the pr, I think - it's good.
1) Maybe you could adjust API with some syntactic sugar, e.g. implement Replacer as a mutable sequence datatype?  (Take look on the Transform class.)

Something like this:
```
>>> replacer[Equality(3*x + 4, 0)]
-4/3
>>> replacer.append({Equality(a_*x, b_): b_/a_})
>>> replacer[Eq(3*x, 4)]
4/3
```
2) I also think, that either `clean()` call should be implicit (e.g. in `append()`, `__setitem__()` and so on) - or there must be a public API for this.

PS: Maybe later, when I did release, I'll take closer look on your pr.

Upabjojr:
>     1. Maybe you could adjust API with some syntactic sugar, e.g. implement Replacer as a mutable sequence datatype?  (Take look on the Transform class.)

Sounds interesting, but I would prefer to keep the API simple right now. Maybe that could be an idea for a future extension.

Mathematica solves the problem of the commutativity of nodes by distinguishing between _Times_ and _Dot_. SymPy's idea of mixing them both into _Mul_ was not optimal.

skirpichev:
@Upabjojr, I've seen you ping me in https://github.com/sympy/sympy/pull/20794#issuecomment-760153302 .  Perhaps, it's a little outdated to discuss that pr ;)  Yet, please keep in mind next time, that I'm blocked in the sympy repos and not able to receive notifications.

PR #540, opened 2017-09-17
[poc] Implement exp as a subclass of Pow

skirpichev:
Quick POC implementation, suggested in [this thread](https://github.com/sympy/sympy/pull/13211#pullrequestreview-58880127).

@Upabjojr, you may be interested, but for me this looks too fragile.

ToDo
--------------

- [ ] make is Function too?
- [ ] avoid arguments hack
- [ ] sqrt
- [ ] cbrt

### DIFF
--- a/diofant/functions/elementary/exponential.py
+++ b/diofant/functions/elementary/exponential.py
@@ -138,7 +138,7 @@ class exp_polar(Function):
         return self.func(1), Mul(*self.args)
 
 
-def exp(arg, **kwargs):
+class exp(Pow):
     """
     The exponential function, `e^x`.
 
@@ -147,7 +147,37 @@ def exp(arg, **kwargs):
 
     diofant.functions.elementary.exponential.log
     """
-    return Pow(S.Exp1, arg, **kwargs)
+
+    def __new__(cls, arg, *opt_args, **kwargs):
+        args = [S.Exp1, arg]
+        if opt_args:
+            args[-1] = opt_args[0]
+        obj = super(exp, cls).__new__(cls, *args, **kwargs)
+        if obj.is_Pow and obj.base is S.Exp1:
+            obj._args = (sympify(args[-1]),)
+        return obj
+
+    @property
+    def base(self):
+        return S.Exp1
+
+    @property
+    def exp(self):
+        return self.args[0]
+
+    def _eval_rewrite_as_sin(self, exp):
+        from .. import sin
+        I = S.ImaginaryUnit
+        return sin(I*self.exp + S.Pi/2) - I*sin(I*self.exp)
+
+    def _eval_rewrite_as_cos(self, exp):
+        from .. import cos
+        I = S.ImaginaryUnit
+        return cos(I*self.exp) + I*cos(I*self.exp + S.Pi/2)
+
+    def _eval_rewrite_as_tanh(self, exp):
+        from .. import tanh
+        return (1 + tanh(self.exp/2))/(1 - tanh(self.exp/2))
 
 
 class log(Function):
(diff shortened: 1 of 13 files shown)


Upabjojr:
This actually looks like a good idea, a better alternative than using the metaclass.

skirpichev:
I don't like used here hacks.  And I like see exp to be Function too.

diofant/functions/elementary/exponential.py
```diff
@@ -147,7 +147,37 @@ def exp(arg, **kwargs):
 
     diofant.functions.elementary.exponential.log
     """
-    return Pow(S.Exp1, arg, **kwargs)
+
+    def __new__(cls, arg, *opt_args, **kwargs):
```
Upabjojr: What about `arg1, arg2=None`? If arg2 is given, then arg1 has to be Exp1. Otherwise arg1 becomes arg2 and arg1 is Exp1.
```
if arg2 is not None:
    assert arg1 == Exp1
else:
    arg2 = arg1
    arg1 = Exp1
```

skirpichev: This is still the same hack, isn't?

What I actually would like to see - strictly 1-arg function exp (and similar for sqrt and cbrt, btw).

Upabjojr: Well, it might be easier with sqrt as the argument has the same position as in Pow. I believe that position should be preserved, otherwise some inherited methods relying on the position of arguments may cause problems. How will the pattern marcher work on exp inheriting Pow?

If you really want to define such functions, I would consider supporting something similar to Haskell-like currying.

skirpichev: > otherwise some inherited methods relying on the position of arguments

Yes, the major reason for implemented hack with arguments - some Pow's methods use "subclassing-friendly" (tm) way of rebuilding expression, i.e. ``self.func(*self.args)``.

> How will the pattern marcher work on exp inheriting Pow?

What you mean?

```
In [1]:In [1]: x_, y_ = symbols('x,y', cls=Wild)

In [2]: exp(x).match(y_**x_)
Out[2]: {x_: x, y_: ℯ}
```

Upabjojr: I believe we need a way of creating a new function by to reducing the number of arguments in a function. Python would kind of allow this by defining `exp = lambda x: Pow(Exp1, x)`. This creates a lambda, but the ideal is to have something similar that creates a Function object by partly defining some arguments.

Or maybe allowing Pow to support a single argument call: Pow(Exp1) would create exp, so that Pow(Exp1)(x) == Pow(Exp1, x).

skirpichev: > I believe we need a way of creating a new function by to reducing the number of arguments in a function.

And how this will fix Pow's methods, that use "subclassing-frienldy" (tm) way of re-creating expression, namely ``self.func(b, e)``?

> Pow(Exp1) would create exp

Functions are not "first class citizens" in diofant (or sympy), e.g. you can't do ``Add(Function('f'), Function('g'))``.  So, doesn't make sense for me.

Upabjojr: > Functions are not "first class citizens" in diofant (or sympy), e.g. you can't do Add(Function('f'), Function('g')). So, doesn't make sense for me.

What's your stance on that? Do you support making functions first-class citizens? There's been a long issue about that in SymPy.

skirpichev: I'm just saying that mentioned your suggestion doesn't make sense without this.

skirpichev:
After some more experimentation, I think that this is not going to be working.  At least, not without wiping "subclassing-friendly" recreation of objects with ``self.func(*args)`` entirely from ``Pow`` codebase...

There is definitely something to think of, but at least - current master's version is consistent (``exp`` behave just as ``sqrt``).  Also, it matches Mathematica behavior, where ``Sqrt[x]`` is ``Power[x, 1/2]`` and ``Exp[x]`` is ``Power[E, x]`` and so on.

Upabjojr:
I agree, I would avoid using subclasses unless extremely necessary.

PR #574, opened 2017-11-09
Misc fixes

skirpichev:
- [x] correct rest of as_real_imag() methods


jksuom:
Note that the variants of erf (and fresnels) also have the same issue.

skirpichev:
besel.py affected too.

PR #612, opened 2018-02-19
[wip] Fix disjoint criteria for complex roots

skirpichev:
@smichr, I think that your patch 73fafba was not the right thing for problem it address.  Instead, we should keep refinement process, until real parts will be disjointed.

Key ingredient to terminate this process: we should consider case when tested pair of roots are on the boundary (western edge) of an isolation rectangles.

I don't entirely sure that your solution is wrong, but it clearly does some additional assumptions about sorted roots, e.g. that they come as conjugated pairs (which is not true for polynomials over gaussian rationals).

What do you think?

### DIFF
--- a/diofant/polys/rootisolation.py
+++ b/diofant/polys/rootisolation.py
@@ -1840,12 +1840,10 @@ class ComplexInterval:
             # In this case, both roots could be on western edges.  Lets
             # test that and terminate refinement if it's the case.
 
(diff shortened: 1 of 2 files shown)


skirpichev:
Ok, @smichr, now I sure that your solution is wrong, see https://github.com/sympy/sympy/issues/14291.

Unfortunately, this approach also don't catch this situation.

PR #630, opened 2018-03-16
Root isolation for polynomials with algebraic coefficients

skirpichev:
- [x] ~~detect cm fields~~, see https://github.com/sagemath/sage/commit/ba389ae418ece8a3e1f8d0e75ebd7cad8e59baca
- [x] special class for real algebraic fields
- [x] alter construct_domain to make real algebraic ground domains where needed
- [x] comparisons without mpmath
- [x] better representation for gaussian rationals and friends (Complex type?)

see https://groups.google.com/d/msg/sympy/Ujznd13xfgw/LPaqaWnbr40J

### DIFF
--- a/diofant/domains/__init__.py
+++ b/diofant/domains/__init__.py
@@ -15,7 +15,7 @@ from .integerring import GMPYIntegerRing, IntegerRing, PythonIntegerRing  # noqa
 from .rationalfield import GMPYRationalField, PythonRationalField, RationalField  # noqa: F401
 from .realfield import RealField
 from .complexfield import ComplexField
-from .algebraicfield import AlgebraicField, RealAlgebraicField  # noqa: F401
+from .algebraicfield import AlgebraicField, ComplexAlgebraicField, RealAlgebraicField  # noqa: F401
 from .expressiondomain import ExpressionDomain
 from .groundtypes import PythonRational  # noqa: F401
 
--- a/diofant/domains/algebraicfield.py
+++ b/diofant/domains/algebraicfield.py
@@ -3,7 +3,7 @@
 import functools
 import operator
 
-from ..core import Integer, sympify
+from ..core import I, Integer, sympify
 from ..core.sympify import CantSympify
 from ..polys.densearith import (dmp_neg, dmp_pow, dmp_rem, dup_add, dup_mul,
                                 dup_sub)
@@ -15,6 +15,7 @@ from ..printing.defaults import DefaultPrinting
 from .characteristiczero import CharacteristicZero
 from .domainelement import DomainElement
 from .field import Field
+from .rationalfield import RationalField
 from .simpledomain import SimpleDomain
 
 
@@ -56,7 +57,14 @@ class AlgebraicField(Field, CharacteristicZero, SimpleDomain):
         else:
             ext_root = None
 
-        obj = super().__new__(RealAlgebraicField if is_real else cls)
+        if is_real:
+            new_cls = RealAlgebraicField
+        elif isinstance(dom, (RationalField, RealAlgebraicField)) and ext == I:
+            new_cls = ComplexAlgebraicField
+        else:
+            new_cls = cls
+        obj = super().__new__(new_cls)
+
         obj.ext = ext
         obj._ext_root = ext_root
         obj.minpoly = minpoly
@@ -69,7 +77,12 @@ class AlgebraicField(Field, CharacteristicZero, SimpleDomain):
         try:
             obj.dtype = _algebraic_numbers_cache[(obj.domain, obj.ext)]
         except KeyError:
-            dtype_cls = RealAlgebraicElement if is_real else AlgebraicElement
+            if new_cls == RealAlgebraicField:
+                dtype_cls = RealAlgebraicElement
+            elif new_cls == ComplexAlgebraicField:
+                dtype_cls = ComplexAlgebraicElement
+            else:
+                dtype_cls = AlgebraicElement
             obj.dtype = type(dtype_cls.__name__, (dtype_cls,), {"_parent": obj})
             _algebraic_numbers_cache[(obj.domain, obj.ext)] = obj.dtype
 
@@ -193,6 +206,12 @@ class RealAlgebraicField(AlgebraicField):
     def is_negative(self, a):
         """Returns True if ``a`` is negative. """
         return a < 0
+
+
+class ComplexAlgebraicField(AlgebraicField):
+    """A class for representing complex algebraic number fields. """
+
+    is_ComplexAlgebraicField = True
 
 
 class AlgebraicElement(DomainElement, CantSympify, DefaultPrinting):
@@ -343,7 +362,25 @@ class AlgebraicElement(DomainElement, CantSympify, DefaultPrinting):
                                                   for _ in self.rep), ZZ.one))
 
 
-class RealAlgebraicElement(AlgebraicElement):
+class ComplexAlgebraicElement(AlgebraicElement):
+    """Elements of complex algebraic numbers field. """
+
+    @property
+    def real(self):
+        """Returns real part of ``self``. """
+        return self.domain.convert(self.rep[-1]) if self else self.domain.zero
+
+    @property
+    def imag(self):
+        """Returns imaginary part of ``self``. """
+        return self.domain.convert((self - self.real)/self.parent.unit)
+
+    def conjugate(self):
+        """Returns the complex conjugate of ``self``. """
+        return self.real - self.parent.unit*self.imag
+
+
+class RealAlgebraicElement(ComplexAlgebraicElement):
     """Elements of real algebraic numbers field. """
 
     def __abs__(self):
@@ -378,3 +415,13 @@ class RealAlgebraicElement(AlgebraicElement):
 
     def __ge__(self, other):
         return self._cmp(other, operator.ge)
+
+    @property
+    def real(self):
+        """Returns real part of ``self``. """
+        return self
+
+    @property
+    def imag(self):
+        """Returns imaginary part of ``self``. """
+        return self.parent.zero
--- a/diofant/domains/domain.py
+++ b/diofant/domains/domain.py
@@ -35,6 +35,7 @@ class Domain(DefaultPrinting, abc.ABC):
     is_ComplexField = is_CC = False
     is_AlgebraicField = is_Algebraic = False
     is_RealAlgebraicField = False
+    is_ComplexAlgebraicField = False
     is_PolynomialRing = is_Poly = False
     is_FractionField = is_Frac = False
     is_SymbolicDomain = is_EX = False
--- a/diofant/polys/compatibility.py
+++ b/diofant/polys/compatibility.py
@@ -281,7 +281,7 @@ class IPolys:
     def dup_real_imag(self, f):
         ring = self
         p, q = dup_real_imag(ring.wrap(f).drop(1).to_dense(), ring.domain)
-        if ring.domain.is_AlgebraicField:
+        if ring.domain.is_ComplexAlgebraicField:
             ring = ring.to_ground()
         return ring.from_dense(p), ring.from_dense(q)
 
--- a/diofant/polys/densetools.py
+++ b/diofant/polys/densetools.py
@@ -2,7 +2,6 @@
 
 import math
 
-from ..core import I
 from ..utilities import variations
 from .densearith import (dmp_add, dmp_add_term, dmp_div, dmp_expand,
                          dmp_exquo_ground, dmp_mul, dmp_mul_ground, dmp_neg,
@@ -579,12 +578,12 @@ def dup_real_imag(f, K):
     >>> R.dup_real_imag(x**3 + x**2 + x + 1)
     (x**3 + x**2 - 3*x*y**2 + x - y**2 + 1, 3*x**2*y + 2*x*y - y**3 + y)
     """
-    if K.is_AlgebraicField and K.ext.as_expr() == I:
+    if K.is_ComplexAlgebraicField:
         K0 = K.domain
-        r1, i1 = dup_real_imag([_.to_dict().get((0,), K0.zero) for _ in f], K0)
-        r2, i2 = dup_real_imag([_.to_dict().get((1,), K0.zero) for _ in f], K0)
+        r1, i1 = dup_real_imag([_.real for _ in f], K0)
+        r2, i2 = dup_real_imag([_.imag for _ in f], K0)
         return dmp_add(r1, dmp_neg(i2, 1, K0), 1, K0), dmp_add(r2, i1, 1, K0)
-    elif not K.is_IntegerRing and not K.is_RationalField:
+    elif not K.is_IntegerRing and not K.is_RationalField and not K.is_RealAlgebraicField:
         raise DomainError("computing real and imaginary parts is not supported over %s" % K)
 
     f1 = dmp_zero(1)
--- a/diofant/polys/rootisolation.py
+++ b/diofant/polys/rootisolation.py
@@ -1526,11 +1526,10 @@ def dup_isolate_complex_roots_sqf(f, K, eps=None, inf=None, sup=None, blackbox=F
                     _roots.append(croot)
                 _roots.append(root)
             roots = _roots
-        elif F.is_AlgebraicField and F.ext.as_expr() == I:
+        elif F.is_ComplexAlgebraicField:
             # Take conjugated polynomial to get solutions in the
             # bottom half-plane.
-            f = [F([-_.to_dict().get((1,), F.domain.zero),
-                    +_.to_dict().get((0,), F.domain.zero)]) for _ in f]
+            f = [_.conjugate() for _ in f]
             roots += [_.conjugate()
                       for _ in dup_isolate_complex_roots_sqf(f, F, eps=eps,
                                                              inf=(u, 0), sup=(s, -v),
@@ -1834,8 +1833,7 @@ class ComplexInterval:
                 f = dmp_add(f1, dmp_mul_ground(f2, dom.unit, 0, dom), 0, dom)
                 f = dmp_compose(f, [-dom.unit, 0], 0, dom)
                 if i.conj:
-                    f = [dom([-_.to_dict().get((1,), dom.domain.zero),
-                              +_.to_dict().get((0,), dom.domain.zero)]) for _ in f]
+                    f = [_.conjugate() for _ in f]
                 f = dmp_compose(f, [dom.unit, 0], 0, dom)
                 r = dup_isolate_real_roots(f, dom, inf=i.ay, sup=i.by)
                 if len([1 for _ in r if ((not i.conj and i.ay < _[0][0]) or
--- a/diofant/polys/rootisolation.py
+++ b/diofant/polys/rootisolation.py
@@ -1243,7 +1243,7 @@ def dup_count_complex_roots(f, K, inf=None, sup=None, exclude=None):
 
     f1, f2 = dup_real_imag(f, F)
 
-    if F.is_AlgebraicField:
+    if F.is_ComplexAlgebraicField:
         F = F.domain
 
     return _count_roots(f1, f2, F, (u, v), (s, t), exclude)
@@ -1542,7 +1542,7 @@ def dup_isolate_complex_roots_sqf(f, K, eps=None, inf=None, sup=None, blackbox=F
 
     f1, f2 = dup_real_imag(f, F)
 
-    if F.is_AlgebraicField:
+    if F.is_ComplexAlgebraicField:
         F = F.domain
 
     f1L1 = dmp_eval_in(f1, v, 1, 1, F)
--- a/diofant/polys/rootoftools.py
+++ b/diofant/polys/rootoftools.py
@@ -121,7 +121,7 @@ class RootOf(Expr):
 
         coeff, poly = preprocess_roots(poly, extension=extension)
 
-        if poly.domain.is_IntegerRing or poly.domain == QQ.algebraic_field(I):
+        if poly.domain.is_IntegerRing or poly.domain.is_AlgebraicField:
             root = cls._indexed_root(poly, index)
         else:
             root = poly, index
@@ -436,7 +436,7 @@ class RootOf(Expr):
     def _all_roots(cls, poly):
         """Get real and complex roots of a composite polynomial. """
 
-        if not poly.domain.is_IntegerRing and poly.domain != QQ.algebraic_field(I):
+        if not (poly.domain.is_IntegerRing or poly.domain.is_AlgebraicField):
             return [(poly, i) for i in range(poly.degree())]
 
         _, factors = poly.factor_list()
--- a/diofant/polys/tests/test_densetools.py
+++ b/diofant/polys/tests/test_densetools.py
@@ -2,7 +2,7 @@
 
 import pytest
 
-from diofant import I, Rational, sin
+from diofant import I, Rational, sin, sqrt
 from diofant.abc import x
 from diofant.domains import EX, FF, QQ, ZZ
 from diofant.polys.densearith import dmp_mul_ground
@@ -431,6 +431,21 @@ def test_dup_real_imag():
     i = 4*x**3*y + x**3 - 4*x*y**3 - 3*x*y**2 - y
 
     assert R.dup_real_imag(f) == (r, i)
+
+    K = ZZ.algebraic_field(sqrt(2))
+    R = K.poly_ring("x", "y")
+    x, y = R.gens
+
+    f = R.x**2 + sqrt(2)*R.x - 1
+    assert R.dup_real_imag(f) == (x**2 - y**2 + sqrt(2)*x - 1, 2*x*y + sqrt(2)*y)
+
+    K = ZZ.algebraic_field(sqrt(2)).algebraic_field(I)
+    R = K.poly_ring("x", "y")
+    x, y = R.to_ground().gens
+
+    f = R.x**2 + 2*sqrt(2)*I*R.x - 1 + I
+    assert R.dup_real_imag(f) == (x**2 - y**2 - 2*sqrt(2)*y - 1,
+                                  2*x*y + 2*sqrt(2)*x + 1)
 
 
 def test_dup_mirror():
--- a/diofant/polys/tests/test_rootisolation.py
+++ b/diofant/polys/tests/test_rootisolation.py
@@ -767,6 +767,22 @@ def test_dup_count_complex_roots_8():
     assert R.dup_count_complex_roots(f, c, d) == 4
 
 
+def test_dup_count_complex_roots_9():
+    R, x = ring("x", QQ.algebraic_field(sqrt(2)))
+
+    f = -x**3 + sqrt(2)*x - 1
+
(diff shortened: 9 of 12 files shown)


skirpichev:
@jksuom, you may want to look on this.  If you feel this can be helpful for sympy, I will appreciate review.  @cswiercz, this also solves problem in sympy/sympy#9366:
```
In [1]: aroot = RootOf(x**7 + x + 1, 0, radicals=False)

In [2]: RootOf(y**2 - aroot, 0, radicals=False, extension=True)
Out[2]: 
      ⎛ 2         ⎛ 7           ⎞   ⎞
RootOf⎝y  - RootOf⎝x  + x + 1, 0⎠, 0⎠

In [3]: _.n()
Out[3]: 2.2937127585107e-28 - 0.892493335621313⋅ⅈ
```

As you can see, there is no changes in rootisolation.py logic. But so far, current code is slow for algebraic domains, take look on the comparison with several rational approximations for coefficients:
```
In [1]: %time RootOf(x**4 + sqrt(2)*x**3 - I*x + 1, 0, extension=True)
CPU times: user 18 s, sys: 96 ms, total: 18.1 s
Wall time: 18.1 s
Out[1]: 
      ⎛ 4     ___  3             ⎞
RootOf⎝x  + ╲╱ 2 ⋅x  - ⅈ⋅x + 1, 0⎠

In [2]: %time RootOf(x**4 + 46341*x**3/32768 - I*x + 1, 0, extension=True)
CPU times: user 664 ms, sys: 0 ns, total: 664 ms
Wall time: 664 ms
Out[2]: 
      ⎛            3             ⎞
      ⎜ 4   46341⋅x              ⎟
RootOf⎜x  + ──────── - ⅈ⋅x + 1, 0⎟
      ⎝      32768               ⎠

In [3]: %time RootOf(x**4 + 6369051672525773*x**3/4503599627370496 - I*x + 1, 0, extension=True)
CPU times: user 792 ms, sys: 4 ms, total: 796 ms
Wall time: 803 ms
Out[3]: 
      ⎛                       3             ⎞
      ⎜ 4   6369051672525773⋅x              ⎟
RootOf⎜x  + ─────────────────── - ⅈ⋅x + 1, 0⎟
      ⎝       4503599627370496              ⎠

```

dup_root_upper_bound() wasn't modified yet to return bound for algebraic domains, but it seems to be not the real reason for this slowdown.

jksuom:
Thanks, this looks interesting. I have only been reading the code, so I cannot see exactly where the time might be spent. There are a couple of points that probably should be investigated. The new `_construct_algebraic` takes the real and imaginary parts. Will those appear in the unevaluated form `(expr + conjugate(expr))/2`? Also, what happens when the sign of a real algebraic number is computed?

skirpichev:
> so I cannot see exactly where the time might be spent.

I can guess: maybe it's comparisons, or factorization (which is noticeably slower for algebraic domains).

> The new _construct_algebraic takes the real and imaginary parts. Will those appear in the unevaluated form (expr + conjugate(expr))/2?

It might be, but I don't consider this as a principal problem.  If ``minimal_polynomial()`` can handle stuff in this unevaluated form - it's fine:
```
In [12]: sqrt(2 + root(2 + I, 7))
Out[12]: 
   _______________
  ╱     7 _______ 
╲╱  2 + ╲╱ 2 + ⅈ  

In [13]: minimal_polynomial(_)(x)
Out[13]: 
 28       26        24         22          20          18           16           14           12            10            8           
x   - 28⋅x   + 364⋅x   - 2912⋅x   + 16016⋅x   - 64064⋅x   + 192192⋅x   - 439300⋅x   + 768824⋅x   - 1025360⋅x   + 1026144⋅x  - 747712⋅x

6           4           2        
  + 375424⋅x  - 116480⋅x  + 16901
```

> Also, what happens when the sign of a real algebraic number is computed?

@jksuom, probably, you are about comparisons of real algebraic numbers?  You can see how implemented [magic methods](https://github.com/skirpichev/diofant/blob/4e149252b4f2fe8f453665a9c963dce39b99563c/diofant/domains/algebraicfield.py#L316-L325).  They use mpmath capabilites, which may be suboptimal.

jksuom:
> They use mpmath capabilites, which may be suboptimal.

That is a possibility that I also occurred to me.

skirpichev:
Support for unevaluated conjugate added in #661.

jksuom:
That is what I thought, too. (Or, maybe `isinstance(ex, conjugate)`. I'm not sure why `__class__` is favoured here.)

skirpichev:
> That is what I thought, too.

Now it's implemented for groebner-based method as well.  I'm not sure it's optimal.

> Or, maybe isinstance(ex, conjugate). I'm not sure why \_\_class\_\_ is favoured here.

For consistency with the code around.  Which is there for usual reasons in sympy: as a historical quirk or just nobody did review of the added in \_minpoly_compose() code.  I will change all that to isinstance calls, which looks consistent with minpoly_groebner().

BTW, it seems ``conjugate()`` - not the only case where ``minimal_polynomial()`` algorithms aren't robust enough.  Same is for Abs.  I did #662 to track this problem.  Feel free to open issue in sympy, as I can't do this and the problem seems affected sympy as well.

But I think that this problem shouldn't block current pr.

skirpichev:
I'm not very convinced how represented gaussian rationals and, more generically, two-level towers of extensions, which support conjugation (i.e. QQ.algebraic_field(sqrt(2)).algebraic_field(I)).

You [suggested](https://github.com/sympy/sympy/issues/9366#issuecomment-399016014) Complex(A) type (field of pairs of elements of field A).  If I correctly understand, this construction should work for any underlying field A (i.e. rational function field?) and must be idempotent, i.e Complex(Complex(A)) == Complex(A)?  Or do you think it shouldn't be too generic and could be useful only for A == RationalField or A == RealAlgebraicField?

I don't sure that Complex is a good name.

jksuom:
> should work for any underlying field A (i.e. rational function field?) and must be idempotent, i.e Complex(Complex(A)) == Complex(A)?

It would be simple to define this for any (commutative) ring (e.g., Z to get Gaussian integers).  If we want to consider only fields, then  one would have to be able to recognize those fields that contain a square root of -1. Otherwise the extension will not be a field.

The name `Complex` was quickly improvised. I think axiom has `Gaussian`. It might be also useful to define `class ComplexAlgebraicField` whose instances would consist of two instances of `RealAlgebraicField`.

skirpichev:
> I think axiom has Gaussian.

Thank you, I'll look into.

> It might be also useful to define class ComplexAlgebraicField

I thought about this option, but in some sense generic AlgebraicField (whose specialization is RealAlgebraicField) **is** a complex field (in sense it can't be embedded in the real field).  But I'm not sure that any such field can be represented as two-level tower of extensions (RealAlgebraic.algebraic_field(I), which will be ComplexAlgebraicField, if I correctly understood your suggestion).

> whose instances would consist of two instances of RealAlgebraicField.

Hmm.  Making RationalField a specialized version of RealAlgebraicField gain more and more sense.

jksuom:
A ComplexAlgebraicField would be isomorphic to an AlgebraicField but with a different implementation where it would be trivial to construct re and im.

skirpichev:
> A ComplexAlgebraicField would be isomorphic to an AlgebraicField but with a different implementation where it would be trivial to construct re and im.

Hmm, I'm not sure that for **any** AlgebraicField it's possible to construct re and im.  E.g. 
```
In [3]: A = QQ.algebraic_field(sqrt(2) + sqrt(3)*I)
In [4]: I in A
Out[4]: False
```
(Remember, AlgebraicField class - for generic field extensions, not just normal ones.)   Should be there some element, that play role ``I`` instead?  I doubt.

Or you are about ComplexAlgebraicField being isomorphic *some* AlgebraicField's, e.g. special tower of extensions like QQ.algebraic_field(sqrt(2)).algebraic_field(I)? That's true, but I see there a little naming problem, mentioned above: in some sense **any** algebraic number field (so, **any** AlgebraicField instance), **which is not** real algebraic field (RealAlgebraicField) - is **complex**, because it can't be embedded in real line.  Not just this special kind of, like ``QQ<sqrt(2)><I>``.

jksuom:
Not every AlgebraicField would be isomorphic to a ComplexAlgebraicField. Only those fields that contain the complex conjugates of their elements would have that property. But every AlgebraicField can be embedded into a ComplexAlgebraicField.

skirpichev:
Thank you, I'll more think about.  Sorry for my stupidity, but right now this still looks like confusing class naming.

FYI: Axiom has the Complex domain, which does exactly match [your proposal here](https://github.com/diofant/diofant/pull/630#issuecomment-406564741), i.e. pair of elements of commutative ring.  But I don't see a Gaussian domain yet.

skirpichev:
Comparisons without mpmath seems to be a little better.

Before:
```
In [1]: %time RootOf(x**4 + sqrt(2)*x**3 - I*x + 1, 0, extension=True)
CPU times: user 12.5 s, sys: 104 ms, total: 12.6 s
Wall time: 14.4 s
Out[1]: 
      ⎛ 4     ___  3             ⎞
RootOf⎝x  + ╲╱ 2 ⋅x  - ⅈ⋅x + 1, 0⎠

In [2]: %time RootOf(x**4 + 46341*x**3/32768 - I*x + 1, 0, extension=True)
CPU times: user 444 ms, sys: 0 ns, total: 444 ms
Wall time: 834 ms
Out[2]: 
      ⎛            3             ⎞
      ⎜ 4   46341⋅x              ⎟
RootOf⎜x  + ──────── - ⅈ⋅x + 1, 0⎟
      ⎝      32768               ⎠
```
After:
```
In [1]: %time RootOf(x**4 + sqrt(2)*x**3 - I*x + 1, 0, extension=True)
CPU times: user 6.31 s, sys: 24 ms, total: 6.34 s
Wall time: 7.27 s
Out[1]: 
      ⎛ 4     ___  3             ⎞
RootOf⎝x  + ╲╱ 2 ⋅x  - ⅈ⋅x + 1, 0⎠
```

Yet, it's still almost order of magnitude worse then the rational domain case.

sympy/sympy#14804 may affect this, as curently end points for intervals are AlgebraicElements (but realy they are rational).  But I doubt it's so severe.

Another variant is ``dup_root_upper_bound()``.  But for rational example above (In[2]) - it doesn't matter if this function return None or some bound.

jksuom:
> Yet, it's still almost order of magnitude worse then the rational domain case.

Hmm.. that feels disappointing. This is how I would like compare rootisolation over algebraic fields vs. rationals.

Algebraic fields:
    Pro: Polynomials of low degree
    Con: Slow comparisons 

Rational field:
    Pro: Fast comparisons
    Con: Polynomials of high degree, with extra roots that have to be sorted out.

Those timing results seem to indicate that using the rationals for root isolation might be the better choice, despite the overhead of high degree polynomials.

skirpichev:
> Those timing results seem to indicate that using the rationals for root isolation might be the better choice, despite the overhead of high degree polynomials.

Not neecessary. Probably, my code for finding the appropriate polynomial over rational field is far from being optimal (btw, [it uses mpmath](https://github.com/diofant/diofant/blob/e32d7e186a40623535eb05d8b178f7a9a38b86bd/diofant/polys/rootoftools.py#L123-L133)), but here is comparision.

With rational coeffs (on master):
```
In [1]: %time RootOf(x**4 + sqrt(2)*x**3 - I*x + 1, 0, extension=True)
CPU times: user 1min 33s, sys: 92 ms, total: 1min 33s
Wall time: 2min 43s
Out[1]: 
      ⎛ 16      14      12      10       8      4      2       ⎞
RootOf⎝x   - 4⋅x   + 8⋅x   - 6⋅x   + 10⋅x  + 5⋅x  + 2⋅x  + 1, 1⎠
```

And on the current branch:
```
In [1]: %time RootOf(x**4 + sqrt(2)*x**3 - I*x + 1, 0, extension=True)
CPU times: user 6.26 s, sys: 32 ms, total: 6.29 s
Wall time: 7.29 s
Out[1]: 
      ⎛ 4     ___  3             ⎞
RootOf⎝x  + ╲╱ 2 ⋅x  - ⅈ⋅x + 1, 0⎠
```

jksuom:
I think that the [norm](https://github.com/sympy/sympy/blob/master/sympy/polys/polytools.py#L3076) of polynomial could be used. Perhaps you could compare with that.

skirpichev:
@jksuom, unless I misunderstand you, norm computation here would be an alternative for calling ``resultant()``?  But it's not a bottleneck in the above code snipet.  Most of time it works inside of ``all_roots()`` and while testing individual roots with ``evalf()``.

jksuom:
Actually, the workhorse of norm is also resultant. But it does not deal with roots.

skirpichev:
Then I doubt this can help.  As I said, resultant computations are fast.  E.g. for above example:
```
In [1]: %time poly = Poly(x**4 + sqrt(2)*x**3 - I*x + 1, x, extension=True)
CPU times: user 244 ms, sys: 12 ms, total: 256 ms
Wall time: 556 ms

In [2]: x, y = poly.gen, Dummy('y')

In [3]: %time p = sum(Poly(c.rep, y)*x**n for (n,), c in poly.rep.terms()).inject(x)
CPU times: user 80 ms, sys: 0 ns, total: 80 ms
Wall time: 209 ms

In [4]: %time q = poly.domain.minpoly.eval(y)
CPU times: user 8 ms, sys: 0 ns, total: 8 ms
Wall time: 17.5 ms

In [5]: %time PurePoly(resultant(p, q, y), x)
CPU times: user 68 ms, sys: 0 ns, total: 68 ms
Wall time: 204 ms
Out[5]: PurePoly(x**16 - 4*x**14 + 8*x**12 - 6*x**10 + 10*x**8 + 5*x**4 + 2*x**2 + 1, x, domain='ZZ')
```

skirpichev:
```
Ring(Domain)
Field(Ring)
CharacteristicZero(Domain)
SimpleDomain(Domain)
AlgebraicField(Field, CharacteristicZero, SimpleDomain)
RealAlgebraicField(AlgebraicField)  <- implement comparision ops
RationalField(RealAlgebraicField, CharacteristicZero, SimpleDomain)
```
and finally for ``RealAlgebraicField.algebraic_field(I)``
```
ComplexAlgebraicField(AlgebraicField)
```
this domain implements re and im methods.

How this class hierarchy looks?

Perhaps, AlgebraicField was named wrongly (and should be reserved for algebraic closure of rationals) and NumberField is better.  Then, RationalField could be derived from AlgebraicField.

jksuom:
Personally, I think that  `Domain` was not a good name for the base class in SymPy because the term domain (in `to_domain`, for example) is used to mean a commutative ring having a special set of attributes. That can probably be changed but that is currently not important for the working of the module.

I also think that `CharacteristicZero` is an odd kind of class. I wonder if it would suffice to have the attribute `characteristic`, and then one could test `A.characteristic == 0`, or something else.

`SimpleDomain` seems to be used for rings that actually are domains in the sense I referred to above. So it is perhaps more descriptive than the base class name. But I wonder if it could be replaced by an attribute in the same way as `characteristic`.

I agree that `NumberField` would be a better name though I am not sure what `AlgebraicField` should be. `RationalField` could be its subclass.

skirpichev:
In fact, Domain, CharacteristicZero and SimpleDomain were above just for completeness.  I'm worry here for stuff in Field/RationalField/AlgebraicField and derived classes.

> I think that Domain was not a good name for the base class in SymPy because the term domain (in to_domain, for example) is used to mean a commutative ring having a special set of attributes

Probably, this could be fixed by moving appropriate modules/attributes to subclasses.  Domain could have only some logic for coersion (convert/convert_from), maybe unify() also should be here.  Perhaps, in this case it's naming will make sense?

Ring should be renamed to CommutativeRing, I think.

> I wonder if it would suffice to have the attribute characteristic

Does make sense too.

> though I am not sure what AlgebraicField should be.

The algebraic closure of rationals, like [in Sage](http://doc.sagemath.org/html/en/reference/number_fields/sage/rings/qqbar.html#sage.rings.qqbar.AlgebraicField).

jksuom:
> Ring should be renamed to CommutativeRing

I would rather allow noncommutative rings, matrix rings, for example, although most rings are currently commutative.

skirpichev:
> I would rather allow noncommutative rings, matrix rings, for example

It should be possible to add a new base class (just Ring?) for CommutativeRing, derived from Domain.

Certainly, this will make sense at some point. (I hope someday domains module will grow to something like new core, more structure-rich, as was [sympycore](https://github.com/pearu/sympycore.  See also https://github.com/sympy/sympy/wiki/Algebras-in-SymPyCore)

> I guess AlgebraicField could be the algebraic closure of rationals (the field of all AlgebraicNumbers).

Yes.

jksuom:
> It should be possible to add a new base class (just Ring?)

I think that could be a good idea (essentially renaming Domain).

skirpichev:
I've started splitting out some more-or-less finished work from here to specialized PR's.  I'll appreciate review.

Real algebraic fields: #669.

skirpichev:
Support for real algebraic domains in rootisolation.py: #673

skirpichev:
Change default construction of algebraic domains for polys: https://github.com/diofant/diofant/pull/656

skirpichev:
Ok, here now is a final version of this PR.  ComplexAlgebraicField introduced.  I'm not sure this is a long-term variant (I would like to unify somehow RationalField and algebraic numbers), but with this subclass things are much more compact and, I hope, much more clear.  I'll appreciate review.

I don't know why root isolation over algebraic fields is so slow (c.f. rational case), but the root algorithm itself seems working as expected, e.g. the dependence of running time on the polynomial degree is cubic.

jksuom:
> why root isolation over algebraic fields is so slow

Probably comparisons are slow in real algebraic fields. In principle, root isolation could always be done over rationals, only with polynomials of higher degree. But that might still take less time.

skirpichev:
> Probably comparisons are slow in real algebraic fields.

Maybe.  I get some speed bonus after reimplementing comparisons to not use mpmath.  But that's not only possibility.  E.g. factorization is slow over algebraic fields.

> In principle, root isolation could always be done over rationals, only with polynomials of higher degree. But that might still take less time.

Mathematica always convert Root's with algebraic coefficients to Root's over integers.  Maybe I also could do such expansions much faster, but currently this way seems to be slower in the Diofant:
```
In [1]: %time RootOf(x**5 + sqrt(2)*x**3 - I*x + 1, 0, extension=True)
CPU times: user 17.3 s, sys: 104 ms, total: 17.4 s
Wall time: 17.5 s
Out[1]: 
      ⎛ 5     ___  3             ⎞
RootOf⎝x  + ╲╱ 2 ⋅x  - ⅈ⋅x + 1, 0⎠

In [2]: %time expand_func(_1)
CPU times: user 3min 16s, sys: 684 ms, total: 3min 17s
Wall time: 3min 27s
Out[2]: 
      ⎛ 20      16      15      12      11      10      8      7      6      5    4      2       ⎞
RootOf⎝x   - 4⋅x   + 4⋅x   + 6⋅x   - 8⋅x   + 6⋅x   + 4⋅x  + 4⋅x  - 4⋅x  + 4⋅x  + x  + 2⋅x  + 1, 2⎠
```

jksuom:
```
In [20]: p = Poly(x**5 + sqrt(2)*x**3 - I*x + 1, x, extension=True)
In [23]: %time p.norm()
CPU times: user 2.89 ms, sys: 6.61 ms, total: 9.5 ms
Wall time: 6.91 ms
Out[23]: 
Poly(x**20 - 4*x**16 + 4*x**15 + 6*x**12 - 8*x**11 + 6*x**10 + 4*x**8 + 4*x**7
 - 4*x**6 + 4*x**5 + x**4 + 2*x**2 + 1, x, domain='QQ')
```

skirpichev:
As I [already said](https://github.com/diofant/diofant/pull/630#issuecomment-408680927) - resultant computations are fast.  Selection of an appropriate root is slow.

jksuom:
Yes. I was thinking of root isolation only.

skirpichev:
Currently, I use mpmath evaluation in ``_eval_expand_func`` helper, but I doubt that I can improve speed of this by order of magnitude, if I replace mpmath with something else.  Anyway, we must isolate all roots of resultant like ``Out[23]`` (BTW, SymPy sort roots wrongly in this example) and then check roots one-by-one until we found some, which solves original polynomial equation.

skirpichev:
> I was hoping that it would not take such a long time.

Maybe I did something wrongly.

> Maybe the process could be accelerated by starting from a floating point approximation of the root in question.

I was thinking about this (using nroots), but I believe it should be the last step in algorithm optimization.

skirpichev:
Correct sorting is the major bottleneck.  For example,
```python
In [18]: %%time
       : e = x**40 - 15*x**17 - 21*x**3 + 11
       : R = ZZ.poly_ring(x)
       : p = R.from_expr(e)
       : R.dup_isolate_all_roots(p)
       : 
CPU times: user 3min 13s, sys: 2.07 s, total: 3min 15s
Wall time: 5min 30s
```
But `Poly(e, x).all_roots()` take hours (doing sorting)...  Maybe there are some better alternatives to #633?

We might start from the numerical solution (as you noted).  But it seems essential to _not_ validate initial rectangles with the Collins-Krandick algorithm.  (This is slow.)  Following snippet use Rouché theorem (suggested [here](https://github.com/sympy/sympy/issues/4697#issuecomment-36993593)):
```python
In [21]: %%time
       : e=x**40 - 15*x**17 - 21*x**3 + 11
       : p=Poly(e, x)
       : dps=15
       : rc = [_ for _ in p.nroots(dps) if not _.is_real]
       : pc = Poly(e, x, domain=CC)
       : # quick verification
       : ans = []
       : for root in rc:
       :     re, im = root.as_real_imag()
       :     radius = max(map(abs, (re, im)))*10**-dps
       :     delta = CC(re, im)
       :     pc0 = abs(pc.shift(delta))
       :     # |a_1|R > |a_0| + |a_2|R^2 + ... |a_n|R^n
       :     cr0 = Poly(2*pc0.coeff_monomial((1,))*x, domain=CC) - pc0
       :     ans.append(bool(cr0.eval(10*radius)))
       : pprint(all(ans))
       : 
       : 
True
CPU times: user 4.94 s, sys: 24.3 ms, total: 4.97 s
Wall time: 7.49 s
```
(We also must check that all discs are pairwise disjoint.) But this test is numerical and I suspect it may fail (produce a wrong ``True``).

BTW, it seems Mathematica's [Root](https://reference.wolfram.com/language/ref/Root.html) suffers from the same problem of bad sorting (as sympy now or as diofant had in the past).  If you use numerical methods (default) - you may get roots in a wrong order:
```
In[1]:= N[{Root[#^20 - 15 #^17 - 21 #^3 + 11 &, 20, 
   ExactRootIsolation -> True],
    Root[#^20 - 15 #^17 - 21 #^3 + 11 &, 20, 
   ExactRootIsolation -> False]}, 20]

Out[1]= {0.79589789805485811114 + 0.67049962856182151814 I, 
 0.97054056319784504470 + 0.26370857625799271442 I}
```

skirpichev:
The slowest part is - the computation of resultants (to isolate zeros, that share common real part).

jksuom:
Should there be an algorithm for this: `radius = max(map(abs, (re, im)))*10**-dps`?

skirpichev:
On Tue, Mar 26, 2019 at 04:06:30AM -0700, Kalevi Suominen wrote:
>    Should there be an algorithm for this: radius = max(map(abs, (re,
>    im)))*10**-dps?

Well, the idea was that nroots() return complex numbers with
dps accurate decimal digits.  The last accurate digit for real and
imaginary component of the root - determine the rectangle for
the root location.  Then, I used 10*radius, which may be suboptimal.

skirpichev:
In fact, resultants computation is slow in Mathematica too (roughly 1min vs 20min in the Diofant or SymPy per one root in the above example).  Apparently, Mathematica doesn't use resultants to detect common real parts or doesn't detect this situation at all (even with ExactRootIsolation->True), using instead some numerical heuristics.

jksuom:
> complex numbers with dps accurate decimal digits.
Then, I used 10*radius, 

That may suffice in general but not necessarily in those cases where some roots are very close to each other. One might try to find the minimum distance between all pairs of roots (though that could be expensive). In any case, I would check the value of the linear coefficient `a_1`. If that is very small, there could be another root close by.

skirpichev:
On Wed, Mar 27, 2019 at 03:48:55AM -0700, Kalevi Suominen wrote:
>    That may suffice in general but not necessarily in those cases where some
>    roots are very close to each other.

As I said, it was assumed that computed approximations are
pairwise disjoint.

>    One might try to find the minimum
>    distance between all pairs of roots (though that could be expensive).

If not, i.e. computed circles of 10*radius does overlap - we can
increase dps.  Here is a heuristics too: if some roots actually
have common real parts - we should stop increase dps at some point.
For such roots - we may fallback to the slow symbolic algorithm,
using resultants.

>    I would check the value of the linear coefficient a_1. If that
>    is very small, there could be another root close by.

If I didn't misunderstand something, the Rouche theorem in the way it
was used above - checks that.

jksuom:
> the Rouche theorem in the way it was used above - checks that.

Yes, I think that should work. A small value of a_1 will easily break the inequality. `|a_1|R > |a_0| + |a_2|R^2 + ... |a_n|R^n`. In that case, R has to be reduced taking care that the exact root is contained in the disk. (So we must have `R > radius` assuming that the exact root has absolute value less that `radius`.)

skirpichev:
The major problem - that this is still a numerical test.

Maybe we should just avoid the sorting phase at
all cost?  We could distinguish roots just by isolation
intervals (instead of assigning indexes).

skirpichev:
It looks also, that for root isolation in the example from  https://github.com/diofant/diofant/pull/630#issuecomment-476344194 - the factorization is the major reason for the slowdown.

If I change dmp_factor_list() to dmp_sqf_list() in [L530](https://github.com/diofant/diofant/blob/bfd2d7d673968c604134c04cd5199b49c62c9810/diofant/polys/rootisolation.py#L530) - timings are much better:
```
In [1]: e = x**40 - 15*x**17 - 21*x**3 + 11
      : 
      : R = ZZ.poly_ring(x)
      : p = R.from_expr(e)

In [2]: %time complexes = R.dup_isolate_complex_roots_sqf(p)
CPU times: user 19.1 s, sys: 41.9 ms, total: 19.1 s
Wall time: 19.1 s
```
c.f. ~3min in https://github.com/diofant/diofant/pull/630#issuecomment-476344194.

This changeset is a reversion of 6c837c0ea.  Maybe there is a better way to avoid mentioned in that commit problem.  Factorization seems to be too costly.

jksuom:
Square free factorization is relatively cheap. The problem is that different (normalized) polynomials may have common roots so the rootof representation is not unique. That can be solved by taking gcd when comparing roots. That is not expensive but an overhead anyway.

skirpichev:
> Square free factorization is relatively cheap.

Yeah, the point was - it's not used in dup_isolate_real_roots_list(), but real factorization instead.

>  That can be solved by taking gcd when comparing roots.

I would prefer a solution, which is more localized to the dup_isolate_real_roots_list().  Maybe we should extract gcd of polys list and handle that part separately.

skirpichev:
Maybe https://github.com/diofant/diofant/pull/835/commits/3aa9aab46239449955ee99ac58bbc570cc28bf47 from #835 is a good start.  It seems this allow to avoid factorization, when `gcd==1` for `len(polys)==2` (most common case).

PR #633, opened 2018-03-31
Fix sorting of complex roots

skirpichev:


skirpichev:
@jksuom, this is a proof of concept implementation of the robust mathematical sorting for polynomial roots, asked in sympy/sympy#14293.  I hope, it follows your suggestions in the issue thread.

So far, it doesn't break anything in the test suite, except ``test_RootOf_evalf()`` (which now randomly fails without commited workarround).  ``test_RootOf_algebraic_domain()`` (abcent in sympy) - also much more slow now, but caching resultants helps a bit.

Please, review this work if it does make sense for you as a variant for sympy's bugfix.

diofant/polys/rootoftools.py
```diff
@@ -299,9 +303,52 @@ def _complexes_sorted(cls, complexes):
...
+            else:
+                return u.bx <= v.ax or v.bx <= u.ax
+
+        res_cache = {}
+
+        def match_same_real_part(u, v):
+            assert u.domain == v.domain
+            dom = u.domain
+            if tuple(u.f) not in res_cache:
+                res_cache[tuple(u.f)] = dmp_resultant(dmp_permute(u.f1, [1, 0], 1, dom),
```
jksuom: Where are `u.f`, `u.f1`, `u.f2` defined?

skirpichev: @jksuom, they are used (or not, as for ``f``) - in the ComplexInterval class for bookeeping.  ``f`` - original polynomial (AFAIK, not saved in Sympy).  ``f1`` and ``f2`` - its real and imaginary parts.

diofant/polys/rootoftools.py
```diff
@@ -299,9 +303,52 @@ def _complexes_sorted(cls, complexes):
...
+                                                      dmp_permute(v.f2, [1, 0], 1, dom),
+                                                      1, dom)
+            g = dmp_gcd(res_cache[tuple(u.f)], res_cache[tuple(v.f)], 0, dom)
+            a, b = max(u.ax, v.ax), min(u.bx, v.bx)
+            roots = dup_isolate_real_roots(g, dom, inf=a, sup=b)
+            if len(roots) != 1:
+                return False
+            else:
+                a, b = roots[0][0]
+            if a == b:
```
skirpichev: Beware, that part seems to be redundant.  But ``dup_count_complex_roots`` sometimes fails in case, where for x-component ``inf==sup``.

For example:
```
In [4]: dup_count_complex_roots([1, 0, 1], ZZ_python, inf=0, sup=0)
Out[4]: 2

In [5]: dup_count_complex_roots([1, 0, 1], ZZ_python, inf=(0, -1), sup=0)
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
<ipython-input-5-d0b6fcdb6ef6> in <module>()
----> 1 dup_count_complex_roots([1, 0, 1], ZZ_python, inf=(0, -1), sup=0)

~/src/diofant/diofant/polys/rootisolation.py in dup_count_complex_roots(f, K, inf, sup, exclude)
   1303     Q_L4 = _intervals_to_quadrants(I_L4, f1L4F, f2L4F, t, v, F)
   1304 
-> 1305     T = _traverse_quadrants(Q_L1, Q_L2, Q_L3, Q_L4, exclude=exclude)
   1306 
   1307     return _winding_number(T, F)

~/src/diofant/diofant/polys/rootisolation.py in _traverse_quadrants(Q_L1, Q_L2, Q_L3, Q_L4, exclude)
   1181                 rules.append((_rules_ambiguous[qq], corners[(j, i)]))
   1182             else:  # pragma: no cover
-> 1183                 raise NotImplementedError("3 element rule (corner): " + str(qq))
   1184 
   1185         q1, k = Q[0], 1

NotImplementedError: 3 element rule (corner): ('OO', 'OO', 'A1')
```
```

File: diofant/polys/rootoftools.py [hunk 3]
```diff
@@ -299,9 +303,52 @@ def _complexes_sorted(cls, complexes):
...
+            dom = u.domain
+            if tuple(u.f) not in res_cache:
+                res_cache[tuple(u.f)] = dmp_resultant(dmp_permute(u.f1, [1, 0], 1, dom),
+                                                      dmp_permute(u.f2, [1, 0], 1, dom),
+                                                      1, dom)
+            if tuple(v.f) not in res_cache:
+                res_cache[tuple(v.f)] = dmp_resultant(dmp_permute(v.f1, [1, 0], 1, dom),
+                                                      dmp_permute(v.f2, [1, 0], 1, dom),
+                                                      1, dom)
+            g = dmp_gcd(res_cache[tuple(u.f)], res_cache[tuple(v.f)], 0, dom)
```
jksuom: I would have expected `u.f` and `v.f` be the same. Otherwise the real parts of `u` and/or `v` might not necessarily be roots of `g`.

diofant/polys/rootoftools.py
```diff
@@ -299,9 +303,52 @@ def _complexes_sorted(cls, complexes):
...
+            a, b = max(u.ax, v.ax), min(u.bx, v.bx)
```
jksuom: I think that `max` and `min` should be interchanged. There could be a single root in the intersection of the intervals but that could be the wrong real part if there are several roots in the union of the intervals.

skirpichev: But comon roots should be in the intersection of isolation rectangles, isn't?  If there is any, of course.  So, this code refines initial bounds for real part, nothing more.  If there are some roots of ``g`` outside of this interval (namely, in the union, not in intersection) - following code expected to return ``False``.

diofant/polys/rootoftools.py
jksuom: I hadn't been thinking about what should be done that case. It could be possible that the two resultants have a common real root but that is not the real part of one (or any) of the roots. Perhaps the intervals should be refined separately for each root until there is only one real root left. Then I think that gcd would be ok.

diofant/polys/rootoftools.py
jksuom: If it is known at this point that there is a single root in a rectangle, then I think this is ok.

skirpichev: Yes, that should be known after [L327](https://github.com/skirpichev/diofant/blob/8cee0050df3574e22bce22b7769cea8e6747a4a5/diofant/polys/rootoftools.py#L327).  If we got something different (not a single root) - procedure will return ``False`` and refinement will be continued on [L351](https://github.com/skirpichev/diofant/blob/8cee0050df3574e22bce22b7769cea8e6747a4a5/diofant/polys/rootoftools.py#L351).

diofant/polys/rootoftools.py
jksuom: The corners are particularly tricky. They may lead to degenerate intervals. Maybe they should be extracted and treated separately.

skirpichev: I've added a rule, that removes need in this workaround.  Not sure if it's correct, yet.

BTW, it seems that only line segments, which are parallel to imaginary axis - poses problems.  Other cases of degenerate isolation rectangles works ok.

diofant/polys/rootoftools.py
skirpichev: @jksuom, I did some cleanup for helper is_same_real_part(), which (hopefuly) make things more transparent and clear.  Let me know if you are still interested in review.

I think, that your objection "It could be possible that the two resultants have a common real root but that is not the real part of one (or any) of the roots." is not valid.  If it's the case -  is_same_real_part() return False.   If you disagree, could you explain what did I miss?

jksuom: I think that there should be no problem if, for each root u and v, the polynomial (in res_cache) having the real part as a root has only one root in the interval (u.ax, u.bx) resp. (v.ax, v.bx). I wasn't sure if that was guaranteed at this point. It seemed that there could be complex roots with far different imaginary part but very nearly the same real part (in the same interval), and then the gcd might catch a wrong real part. That should not happen if the real roots have been isolated beforehand.

skirpichev: > I wasn't sure if that was guaranteed at this point.

@jksuom, but why this should be guaranteed here?

> and then the gcd might catch a wrong real part.

We test, 1) that  are in the intersection of real parts of the original ``u`` an ``v``.  
Then, if there some root (unique!) - 2) test that both original polynomial have one complex root with real part in the interval gcd_roots[0][0].

If there multple gcd_roots (or none) or original polynomials don't have complex roots (or have too much) - return False.  Why do you think that "wrong real parts", coming with gcd_roots will survive both tests?

jksuom: I agree that if there is a single root of gcd in the interval and each polynomial has a single root in its respective complex interval (as in 2 above), then the result True is correct. But I'm still concerned about  multiple roots. Is is correct to always return False if `len(gcd_roots) > 1`? Could it be possible that the real parts are the same and, in addition, each polynomial has another root with same but slightly different real part that happens to be in the same real interval though the imaginary parts may not be in the (ay, by) intervals of u and v?

skirpichev: > Could it be possible that the real parts are the same and, in addition, each polynomial has another root with same but slightly different real part that happens to be in the same real interval

How this could be?  If there are other pair of roots with slightly different real part - we should continue refinement of u and v (by returning False) and eventually intervals for different multiplets of roots with same real part will be disjoint (as refinement process will shrink intersection of u and v).  That's why I do think that we will have len(gcd_roots) == 1 at some point (if u and v belongs to the multiplet of roots with same real part).

> to be in the same real interval though the imaginary parts may not be in the (ay, by) intervals of u and v

Why we ever should care about roots, outside of intervals for u and v?  is_same_real_part() is a decision procedure: for given isolation rectangles u and v we want to decide if these particular roots have same real part or not (in which case we must continue refinement).

skirpichev: > You do not expect is_same_real_part returning False to mean that the real parts cannot possibly be same.

Yes, False just mean: continue refinement.  This process will continue until roots will be either disjoint by real part or gcd will have exactly one root.  In the later case we will check if both polys have complex root with real part in the interval of gcd's root.  If yes - we have found a pair of roots with same real part and should stop refinement.  If not, we will return False and continue refinement again: eventually real parts also will be different.

> I think that this should work correctly.

So, no issues with math you have found so far?

PS: Workaround for sympy/sympy#14738 simplifies code a lot, so I would like to include the real fix for that issue as well.  But original code for ``a==b`` case from 2c706b8fa6 is sufficiently straightforward (substitute x -> a + I*x and search for real solutions), so review of that part is less important.

diofant/polys/rootisolation.py
```diff
@@ -1860,15 +1862,14 @@ def conjugate(self):
                                self.F1, self.F2, self.f1, self.f2,
                                self.f, self.domain, conj=not self.conj)
 
-    def is_disjoint(self, other):
+    def is_disjoint(self, other, strict=False):
```
skirpichev: strict, probably, a wrong name.  Maybe re_distinct?

diofant/polys/rootoftools.py
jksuom: No math issues. This looks good. The name `is_same_real_part` was misleading me. Maybe it could be something like `confirm_same_real_part`.

skirpichev: > Maybe it could be something like confirm_same_real_part.

Yes, that sounds better.  But I'm thinking about moving this functional from that helper function to the ComplexInterval.is_disjoint() method.  Some option could enable all this.

skirpichev:
Now all logic for testing real parts of roots moved to the ComplexInterval.is_disjoint() method.  It could be enabled by ``re_distinct`` boolean option (see docstring).  I'm a bit unsure about naming of this flag.

@jksuom, let me know if you would like to review part, related to the sympy/sympy#14738.

jksuom:
> part, related to the sympy/sympy#14738.

It seems that this is in another PR. Yes, I think I would like to see that. (Though I'm not sure if I can be helpful within reasonable time. The subject is very low level and rather tedious to work with.)

skirpichev:
> It seems that this is in another PR.

You mean, this belongs to another PR? Yeah, but fix should be very simple one-liner and I prefer to solve this issue, not introduce ugly workaround.  Ok, I'll try to handle this myself.

Any final remarks?  ``re_distinct`` flag really misnamed, through I don't see better option yet.

jksuom:
> this belongs to another PR

Yes, though I understood that you might already have that prepared.

`is_disjoint` seems to return `True` when either the rectangles really are disjoint (except possibly at boundary) or `re_distinct` is `True` and the roots are seen to have the same real part (regardless of whether the rectangles overlap or not). That might be confusing to a user as this is a public method.

skirpichev:
> is_disjoint seems to return True when either the rectangles really are disjoint (except possibly at boundary) or re_distinct is True and the roots are seen to have the same real part 

is_disjoint() works as before (test that isolation rectangles are disjoint on complex plane), if re_distinct is False (default).

If re_distinct is True - is_disjoint() will return True either if we can confirm that roots have different real parts OR that they have common real part.  Otherwise, it will return False.  That's documented in the docstring, but option indeed seems to have confusing name.  ``split_real_part`` maybe?

skirpichev:
@jksuom, thank you for your helpful review.  I think that finally may be merged.

jksuom:
Thank you, though I am not sure that I was able to help.

skirpichev:
> I am not sure that I was able to help.

If I not misunderstood, you said no math issues were found you so far.  That view helps, for sure.  

Naming of options (which now is slightly better, I hope) and fixing sympy/sympy#14738 - less important issues.

skirpichev:
FYI: sorting with conjugated pairs of roots corrected in diofant/diofant#658.

skirpichev:
FYI: #741 fix some performance problems with caching, introduced by correct sorting.

skirpichev:
FYI, @smichr, I still believe you mistakenly closed sympy/sympy#14293.

I don't know how exactly the SymPy code works now.  But even if the refinement process doesn't change
the order of roots - the current ordering is too much depends on the details of the root isolation routines.
I'm pretty sure, that even slight change of the rootisolation module (e.g. [this](https://github.com/diofant/diofant/issues/838) possible optimization) - will break the current sorting "definition".  In other words, root indexing will not be
stable across different SymPy versions.

So, there is no unique ordering of complex roots, but I think that some mathematically defined ordering
rules - should be strongly preferred.

PR #654, opened 2018-06-23
Misc fixes

skirpichev:


skirpichev:
@asmeurer, I don't think 4030e3d should go in sympy, but Rational.p/q properties looks too cryptic.  Why not provide numerator/denominator properties?  This would make Rational and Integer types more compatible with [python numbers hierarchy](https://docs.python.org/3/library/numbers.html#module-numbers).  (In fact, Rational class probably will be fully compatible after this.)

asmeurer:
I agree it should support the ABCs.  It looks like there's an open issue for it https://github.com/sympy/sympy/issues/12134. The `p` and `q` properties should be maintained for backwards compatibility. Either way, they are intended for internal use only, since they return ints, not Integer.

skirpichev:
> Either way, they are intended for internal use only, since they return ints, not Integer.

@asmeurer, people can think that attribute is private if it name starts with underscore.  That's not just a widely adopted convention, Python import system has some additional support for this.

But people can't guess that attribute is private, if its naming doesn't suggest that.  And they will use this, for sure, e.g. [see here](https://github.com/mathics/Mathics/blob/6ce083833ce382b7e5d573ab8f94c17b5a1db371/mathics/core/convert.py#L159-L170).

>  sympy/sympy#12134

Easy to fix?  Oh, I don't think so, given you have ">>" for Implies and so on.

asmeurer:
`>>` isn't currently defined on Integer/Rational. Regardless, I don't see an issue with making the logic operators do bitwise operations on integers, since integers aren't booleans. This would actually give a strong reason to have a clear separation of the two.

PR #669, opened 2018-08-05
Specialized class for real algebraic numbers

skirpichev:
For elements there are define usual mathematical comparison ops (undefined for generic AlgebraicElement).  RealAlgebraicElement support \_\_abs__ magic method.

Also, this pull request implements RootOf.refine method.


jksuom:
This looks good though I did not check the coding details. I had to look more closely into the automatic creation of RealAlgebraicFields but I now think that I understand the way it works. It is surely practical.

skirpichev:
> I had to look more closely into the automatic creation of RealAlgebraicFields but I now think that I understand the way it works.

There is a specialized cache version, like used in rings.py, fields.py and modularinteger.py (so, it's more-or-less consistent across the codebase).  Probably, in future I'll replace this stuff with using standard Diofant's caching abilities (cacheit and so on), but right now it seems more hard then I expected.

I'm more concerned about mathematical stuff in [ext_root](https://github.com/skirpichev/diofant/blob/7d0b748dbafbe2aea41196fa5ea5c6ab7e1d7651/diofant/domains/algebraicfield.py#L159-L169) property (this is a RootOf object, representing the primitive element, ext) and [\_cmp](https://github.com/skirpichev/diofant/blob/7d0b748dbafbe2aea41196fa5ea5c6ab7e1d7651/diofant/domains/algebraicfield.py#L338-L354) helper method (which implements comparison ops).

``ext_root`` implemented for real algebraic fields so far, which meets practical needs.

skirpichev:
An Expr instance, like sqrt(2).

jksuom:
It is hard to see, in general, that an Expr, `ext`, is real (even if it is). `ext.is_real` will often return `None`. I see that you play it safe and define the field as an `AlgebraicField` in that case. That would make it plausible that the comparison on line 167 won't raise an exception as it is only attempted when `self.ext.is_real` is true.

skirpichev:
> ext.is_real will often return None.  I see that you play it safe and define the field as an AlgebraicField in that case. 

Perhaps, I should be more careful here and take into account this ternary logic.  In sense that we should raise an exception in constructor, instead of using AlgebraicField as a fallback for ``None``.

However, It's hard to find a case, where ext.is_real would return ``None``.  Perhaps, the only case would be if minpoly is Poly(x, x), when ``ext`` in fact is a zero.  Maybe I should check this pathological scenario in the AlgebraicField constructor.

The point of ``ext_root`` is to select a single root of minpoly for extension ``ext``.  Another variant for this procedure is testing difference ``ext - r`` for all minpoly roots ``r``.  If that difference can be computed with precision (and is nonzero) - this ``r`` is invalid for our purposes.  But I'm afraid that the only way to test this in infallible way - compute ``minpoly(ext - r)``.

jksuom:
>  It's hard to find a case, where ext.is_real would return None. 

They are not quite trivial to find, but the roots of a cubic polynomial tend to be such in the Casus Irreducibilis case. This is one example that I found in `test_roots_cubic()`: 
`eq = -x**3 + 2*x**2 + 3*x - 2`.

skirpichev:
> This is one example that I found in test_roots_cubic(): eq = -x\*\*3 + 2\*x\*\*2 + 3\*x - 2.

Indeed.  But provided above variant with difference testing seems to be working here:
```
In [1]: eq = -x**3 + 2*x**2 + 3*x - 2

In [2]: rs = roots(eq, multiple=True)

In [3]: rs[0].is_real

In [4]: (rs[0] - RootOf(eq, 0)).n(2)
Out[4]: 4.2 - 0.e-7⋅ⅈ

In [5]: (rs[0] - RootOf(eq, 1)).n(2)
Out[5]: 2.3 - 0.e-7⋅ⅈ

In [6]: (rs[0] - RootOf(eq, 2)).n(2)
---------------------------------------------------------------------------
PrecisionExhausted                        Traceback (most recent call last)
...
```

skirpichev:
I used more careful testing of is_real with fallback in case ``ext.is_real is None`` to the full-blown minimal_polynomial-based method, which determine an appropriate root of minpoly.

jksuom:
> if not minimal_polynomial(ext - r)(0):

In which form will `r` be here?

skirpichev:
A RootOf instance.  Well, unless minpoly is linear.  But I think that case
out of the question currently.

skirpichev:
I'm going to merge this, unless you have better ideas how to select an appropriate minpoly root for the primitive element.

jksuom:
I cannot think of a way to avoid numerical evaluation, in one way or another.  There may be several ways of organizing the selection, but this is probably as good as any. Perhaps practical use will bring forth new ideas.

Issue #425, opened 2017-02-06
Wrong parametric limits, like limitinf(x*y, x)


skirpichev:
```python
In [9]: from diofant.series.gruntz import limitinf

In [10]: x = Symbol('x', real=True, positive=True)

In [11]: y = Symbol('y')

In [12]: limitinf(x*y, x)
Out[12]: ∞⋅sign(y)
```

``Piecewise[(0, Eq(y, 0)), (sign(y)*oo, True))`` seems to be more correct.  Also, probably we should replace in ``mrv()`` workaround ``elif any(a.is_infinite for a in Mul.make_args(limitinf(e.exp, x)))`` with correct test ``elif limitinf(e.exp, x).is_infinite`` from the thesis.

skirpichev:
This may require better assumptions system from the core.  E.g. to assume certain properties on expressions, not just atoms.

jldohmann:
Hi, I'm new to development in general and am interested in contributing...at some point after I've looked around and asked enough questions (still learning). 

Let me see if my understanding of the issue is correct: you want `limitinf` to return either all possible conditions similar to `Piecewise` when `y` has no assumptions, or you want a default assumption within the expression (presumably that `y>0`) when there's none specified for `y` ? 

Could an optional third assumptions argument be added to `limitinf`, similar to how Mathematica handles it?

`In [1]: Limit[x*y, x -> Infinity, Assumptions -> {x > 0, y > 0}]`
`Out[1]: ∞`

Apologies if this is old and not on the radar anymore. Just trying to look for stuff I'm interested in and get my feet wet.

skirpichev:
> you want limitinf to return either all possible conditions similar to Piecewise when y has no assumptions

Yes, exactly.

Unfortunately, this is not so simple.  Probably, assumptions handling should be improved first.  E.g. there should be a way to assume something about expressions (not just symbols).

> Could an optional third assumptions argument be added to limitinf, similar to how Mathematica handles it?

Maybe.  But again, this will require a different assumptions system.

> Apologies if this is old and not on the radar anymore.

If the bug is not closed - it's still relevant.  No need to apologize.

skirpichev:
[...]

PR #685, opened 2018-10-04
Misc fixes

skirpichev:


skirpichev:
@jksuom, as you can see, this issue add regression test for https://github.com/sympy/sympy/issues/15323.  I believe, most limit issues for sympy (except parametric limits like #425) are fixed in the diofant more or less on the way you suggested [here](https://github.com/sympy/sympy/issues/15323#issuecomment-478639399), i.e. by reimplementing the Gruntz algorithm from scratch and using the evaluate(False) context where appropriate.  Maybe it could be just ported to the SymPy?

jksuom:
It is interesting to see that you have succeeded in implementing the Gruntz algorithm without SubsSets. I have also disliked those as they make debugging more tedious, but I was under the impression that evaluate=False had been tested and could not be used because of some issues like over-complicated expressions. Thank you for showing that it is possible. The code looks much better. It certainly should be ported to SymPy.

PR #687, opened 2018-10-08
Use type() instead of .func

skirpichev:
[SymPy docs](https://docs.sympy.org/latest/tutorial/manipulation.html) says "func is the head of the object. ... Usually it is the same as the class of the object (though there are exceptions to this rule)."  So far, from the codebase I could grep only AppliedPredicate as an exception.  We should investigate if using a dedicated attribute really does make sense.

- [ ] fix type vs func in docstrings and doctests

### DIFF
--- a/docs/tutorial/manipulation.rst
+++ b/docs/tutorial/manipulation.rst
@@ -190,22 +190,21 @@ Walking the Tree
 ================
 
 Let's look at how to dig our way through an expression tree, using a very
-generic interface --- attributes :attr:`~diofant.core.basic.Basic.func` and
+generic interface --- builtin :class:`type` and attribute
 :attr:`~diofant.core.basic.Basic.args`.
 
-The head of the object is encoded in the :attr:`~diofant.core.basic.Basic.func`
-attribute
+The class of the object encode the head of expression
 
     >>> expr = 2 + x*y; expr
     x⋅y + 2
-    >>> expr.func
+    >>> type(expr)
     <class 'diofant.core.add.Add'>
 
 The class of an object need not be the same as the one used to create it
 
     >>> Add(x, x)
     2⋅x
-    >>> _.func
+    >>> type(_)
     <class 'diofant.core.mul.Mul'>
 
 .. note::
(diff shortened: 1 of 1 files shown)


skirpichev:
See also https://github.com/sympy/sympy/issues/4834

skirpichev:
@asmeurer, what do you think?  I'm not sure that breaking ``foo.func == type(foo)`` may be really too helpful to keep such a possibility.

asmeurer:
With type(foo) there's no way for the head of an object to not be a class.

skirpichev:
On Mon, Nov 05, 2018 at 11:20:34AM -0800, Aaron Meurer wrote:
>    With type(foo) there's no way for the head of an object to not be a class.

I understand.  The question is: does this restriction really matter?

asmeurer:
Well you can see how it's useful for AppliedPredicate, because you can use both Q.positive and Q.positive(x) as objects in expressions. Some people have wanted to do the same thing with functions so you can write `sin + cos`, though it hasn't happened. There are other possibilities too like storing non-Basic "args" in the func instead of the args. Or you could store other things in there without the need to have them in `args` or to have a different class for each combination, like boolean flags on objects or maybe even assumptions (I'm not sure).

Ronan suggested other refactorings could be possible in the issue, too, though I'm not quite sure what he had in mind.

skirpichev:
On Mon, Nov 05, 2018 at 11:20:22PM -0800, Aaron Meurer wrote:
>    Well you can see how it's useful for AppliedPredicate, because you can use
>    both Q.positive and Q.positive(x) as objects in expressions.

I'm not sure, that this case is a good argument in favor of non-trivial
func property.  I saw this example, it was mentioned in the pull request.

"positive" may be a class instance, which overloads __call__ to produce
another class instance for "positive(x)", something like "Application(positive, x)".
Same should be doable for functions (in core).

>    There are other possibilities too like
>    storing non-Basic "args" in the func instead of the args.
>    Or you could
>    store other things in there without the need to have them in args or to
>    have a different class for each combination, like boolean flags on objects
>    or maybe even assumptions (I'm not sure).

Yes, making func to be anything open possibilities to do very different
and very cubersome things.

But I believe there are ways to do same differently.  For example, maybe
for storing flags the current func-args model could be extended to
include also boolean kwargs.

(A propos, looking on the Basic class, especially in SymPy, I see how
it's overbloated with stuff, that actually doesn't make sense here.
In my view, the Basic should implement a bare minimum for working with
tree-like structures: creating, walking on, substitution/replacement,
pattern-matching.  But you have is_hypergeometric, is_comparable,
as_content_primitive...  I like MatchPy approach
here, which allows builtin types and containers.  Maybe one
base class is not a right thing at all.)

skirpichev:
Postponed.

PR #693, opened 2018-10-13
[wip] Alter last clause of Modified BSD License

skirpichev:


skirpichev:
@certik, what do you think about changing last clause of the SymPy license like this to match https://choosealicense.com/licenses/bsd-3-clause/?

This will make the product license detectable for Github, like in [this](https://github.com/skirpichev/gmpy_ctypes), see LGPL-3.0 on top of the repository.  You also may filter by license type while searching.

PR #694, opened 2018-10-13
Support gmpy2.mpz ground type for Rational

skirpichev:


skirpichev:
@isuruf (it seems, you proposed this for sympy in https://github.com/sympy/sympy/issues/10928), looks like there is no significant timing changes for the test suite.  Not sure if it worth.

PR #705, opened 2018-11-01
Simplify setup.py

skirpichev:


skirpichev:
Note, that references to other extras aren't supported, see pypa/setuptools#1260.

PR #710, opened 2018-11-09
Misc fixes

skirpichev:


skirpichev:
@jksuom, I think that https://github.com/sympy/sympy/pull/15540 (you just merged to sympy) has [wrong test](https://github.com/normalhuman/sympy/blob/08d06e74f8a948c19733a4b8f89fa7f81f2d0c5a/sympy/series/tests/test_series.py#L211-L212).

jksuom:
I noticed that and agree. It probably comes from the `n=6` default. In fact, I think that some kind of error should be given by `series()` as `exp` is not meromorphic at infinity.

jksuom:
I should have added a comment to the test to remind that `series()` has to be improved.

BTW, how does Mathematica handle order for series? Are there only powers? Is there also something like `O(exp(x))` or `O(log(x))`?

jksuom:
And how are asymptotic expansions handled? Apparently, there is a large library of ready-made results but are there also methods for deriving new expansions?

skirpichev:
> how does Mathematica handle order for series? Are there only powers? 

There are only ``O[x]`` and ``O[x, x0]``.  I don't think that it can evaluate log(O(x)) or something similar.  Only powers.

> And how are asymptotic expansions handled? 

In the Diofant - used algorithm from Gruntz thesis.  I don't know what Mathematica do, but I doubt that it's just a library of ready results...

PR #712, opened 2018-11-11
[wip] New algorithm for factorization of polynomials over algebraic number fields

skirpichev:
This is a continuation of work in sympy/sympy#2468.

Todo:
- [x] attribute first commit to @katjasophie (mention some trivial changes, like diofant vs sympy imports)
   - [ ] also split out https://github.com/sympy/sympy/commit/b5986e84801a305fafc51827ca40ef010d322925 and https://github.com/sympy/sympy/commit/f52f756b911c529c41c303924542ebf8d0f98f4c?
- [ ] full test coverage
- [ ] better file/functions names?
- [x] integrate with the rest of polys module (polyconfig.py's option?)
- [ ] changes for rref/solve_lin_sys?
- [x] ~~a plan for towers of extensions (important case: ``QQ<real><I>``)?~~
- [ ] fix TODO in file?
- [ ] rebase & cleanup
- [ ] release notes

@katjasophie, I assume, I can use your unmerged yet work from sympy/sympy#2468 to the Diofant on terms of the usual SymPy license.  Please reply, if you are disagreed.

Some benchmarks on my system:
```
In [2]: R, x, y = ring('x, y', QQ.algebraic_field(sqrt(2)))
   ...: f = (x + y)*(x**2 + sqrt(2)*y)
   ...: %timeit efactor(f)
   ...: %timeit f.factor_list()
350 ms ± 3.83 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
226 ms ± 875 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: R, x, y, z = ring('x, y, z', QQ.algebraic_field(root(2, 5)))
   ...: f = (x - y)*(x**2*y + root(2, 5)*y**2*z)
   ...: %timeit efactor(f)
   ...: %timeit f.factor_list()
514 ms ± 12.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
4.44 s ± 545 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [4]: R, x, y, z = ring('x, y, z', QQ.algebraic_field(sqrt(2), root(3, 3)))
   ...: f = (x**2 + 2*y + sqrt(2))*(root(3, 3)*x - z)
   ...: %timeit efactor(f)
   ...: %timeit f.factor_list()
2.66 s ± 141 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
47.4 s ± 2.39 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

New algorithm is clearly better for extensions with high degrees.

### DIFF
--- a/diofant/matrices/matrices.py
+++ b/diofant/matrices/matrices.py
@@ -1,5 +1,5 @@
 import collections
-from functools import reduce
+from functools import partial, reduce
 from types import FunctionType
 
 from ..core import (Add, Atom, Basic, Dummy, Expr, Float, I, Integer, Pow,
@@ -2664,7 +2664,7 @@ class MatrixBase(DefaultPrinting):
 
         return self.adjugate() / d
 
-    def rref(self, iszerofunc=_iszero, simplify=False):
+    def rref(self, iszerofunc=_iszero, simplify=False, scalefunc=None, elimfunc=None):
         """Return reduced row-echelon form of matrix and indices of pivot vars.
 
         To simplify elements before finding nonzero pivots set simplify=True
@@ -2702,12 +2702,14 @@ class MatrixBase(DefaultPrinting):
                 else:
                     continue
             scale = r[pivot, i]
-            r.row_op(pivot, lambda x, _: x / scale)
+            _scalefunc = partial(scalefunc, scale=scale) if scalefunc else lambda x, _: x / scale
+            r.row_op(pivot, _scalefunc)
             for j in range(r.rows):
                 if j == pivot:
                     continue
                 scale = r[j, i]
-                r.zip_row_op(j, pivot, lambda x, y: x - scale*y)
+                _elimfunc = partial(elimfunc, scale=scale) if elimfunc else lambda x, y: x - scale*y
+                r.zip_row_op(j, pivot, _elimfunc)
             pivotlist.append(i)
             pivot += 1
         return self._new(r), pivotlist
--- a/diofant/polys/factorization_alg_field.py
+++ b/diofant/polys/factorization_alg_field.py
@@ -1,0 +1,1038 @@
+import functools
+import operator
+import random
+
+from ..core import Dummy
+from ..domains.algebraicfield import AlgebraicElement
+from ..integrals.heurisch import _symbols
+from ..ntheory import nextprime
+from .modulargcd import (_euclidean_algorithm, _gf_gcdex, _minpoly_from_dense,
+                         _trunc)
+from .polyconfig import using
+from .polyerrors import UnluckyLeadingCoefficient
+from .polyutils import _sort_factors
+from .rings import PolynomialRing
+from .solvers import solve_lin_sys
+
+
+# TODO
+# ====
+
+# -) efficiency of _factor can be improved for irreducible polynomials if the
+#    univariate factorization is done before the LC is factored
+
+
+def _alpha_to_z(f, ring):
+    r"""
+    Change the representation of a polynomial over
+    `\mathbb Q(\alpha)` by replacing the algebraic element `\alpha` by a new
+    variable `z`.
+
+    Parameters
+    ==========
+
+    f : PolyElement
+        polynomial in `\mathbb Q(\alpha)[x_0, \ldots, x_n]`
+    ring : PolynomialRing
+        the polynomial ring `\mathbb Q[x_0, \ldots, x_n, z]`
+
+    Returns
+    =======
+
+    f_ : PolyElement
+        polynomial in `\mathbb Q[x_0, \ldots, x_n, z]`
+
+    """
+    if isinstance(f, AlgebraicElement):
+        ring = ring.drop(*ring.gens[:-1])
+        f_ = ring(dict(f.rep))
+
+    else:
+        f_ = ring.zero
+
+        for monom, coeff in f.items():
+            coeff = coeff.rep.to_dense()
+            n = len(coeff)
+
+            for i in range(n):
+                m = monom + (n-i-1,)
+                if coeff[i]:
+                    f_[m] = coeff[i]
+
+    return f_
+
+
+def _z_to_alpha(f, ring):
+    r"""
+    Change the representation of a polynomial in
+    `\mathbb Q[x_0, \ldots, x_n, z]` by replacing the variable `z` by the
+    algebraic element `\alpha` of the given ring
+    `\mathbb Q(\alpha)[x_0, \ldots, x_n]`.
+
+    Parameters
+    ==========
+
+    f : PolyElement
+        polynomial in `\mathbb Q[x_0, \ldots, x_n, z]`
+    ring : PolynomialRing
+        the polynomial ring `\mathbb Q(\alpha)[x_0, \ldots, x_n]`
+
+    Returns
+    =======
+
+    f_ : PolyElement
+        polynomial in `\mathbb Q(\alpha)[x_0, \ldots, x_n]`
+
+    """
+    domain = ring.domain
+
+    f_ = ring.zero
+    for monom, coeff in f.items():
+        m = monom[:-1]
+        c = domain([domain.domain(coeff)] + [0]*monom[-1])
+
+        if m not in f_:
+            f_[m] = c
+        else:
+            f_[m] += c
+
+    return f_
+
+
+def _distinct_prime_divisors(S, domain):
+    r"""
+    Try to find pairwise coprime divisors of all elements of a given list
+    `S` of integers.
+
+    If this fails, ``None`` is returned.
+
+    References
+    ==========
+
+    * :cite:`Javadi2009factor`
+
+    """
+    gcd = domain.gcd
+    divisors = []
+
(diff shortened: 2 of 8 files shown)


jksuom:
Thanks for pointing this out. I wasn't aware that there was unmerged work on factorization. It looks like there was no-one to review the PR.  I'll need some time to look into this. Then I'll try to return.

jksuom:
I have not inspected everything but this gives a positive impression. The quality of code is good. The algorithm is a simplified version of Javadi-Monaghan but it looks like the essential parts are included.

My only reservations are also in your TODO list. There is a large number of private low level routines with file scope. That makes the code less readable. I would have expected some classes that would clarify the structure and contain attributes that are now repeated on some (long) parameter lists. Also the necessity of changing rref should probably be reconsidered.

skirpichev:
> a simplified version of Javadi-Monaghan

Yeah, unfortunately this doesn't help with univariate case, it handled by the Trager algorithm as before.  This will lover priority of this work, but I still hope to include it in the next release.

> the necessity of changing rref should probably be reconsidered.

I see not too much options there, with the current matrices module.

jksuom:
> it handled by the Trager algorithm as before.

It looks like there may be no other choice. Polynomials over fraction fields of unique factorization domains might be handled directly in the same way as polynomials over Q, but only few algebraic number fields have that property.

skirpichev:
BTW, there are algorithms for polynomial gcd computation over algebraic number fields (modulargcd.py), from same GSoC (and same person).  I think, this can be exposed via PolyElement's gcd() method and would be better than the current fallback method (dmp_inner_gcd, working with dense polys).

skirpichev:
Regarding `rref()` changes, it looks like just a Gauss elimination for matrices with coefficients in Z/nZ, where `n` is a composite integer in general (we start from n`= prime `p` and then square this number on each iteration).  In general, it's not a field and `rref()` may fail.  From the cited paper (see Alg.2), it seems if there will be a failure due to non-invertible factor - it's fine, algorithm should be just restarted with a different `p` (btw, the current code doesn't catch `NonInvertible` exceptions).

I'm not sure what to do.  If I relax assumptions (assert) in `solve_lin_sys()` (i.e. domain of the polynomial ring may no be a field) AND add the FiniteRing domain for the ring of integers modulo `n` - no changes in `rref()` will be required (just before calling of  `solve_lin_sys()` - the ring domain will be changed to Z/nZ).  Does this sound better than current `rref()` changes?

jksuom:
> Does this sound better than current rref() changes?

Yes, it does to me. It seems to me that little or no changes would be needed in `rref()` if `iszerofunc` is defined as `is_invertible` in case of domain Z/nZ.

skirpichev:
Hmm, a "little" change may be required.  `rref()` currently uses `/` (i.e. `__truediv__`) for scaling.  Unfortunately, this is defined for ZZ (both with python and gmpy ground types) and thus for polynomial ring over ZZ - the current form of `rref()` will not make any sense.

I'm not sure how to fix this without a proper support for domain types in the matrices module.  With knowledge about the domain of elements - we could use something like `domain.revert(scale)` to invert scaling factor (if possible).  But with hackish solution as present with the `RawMatrix` class (which just add an override for `Matrix._sympify` helper) - this is impossible.

I know, there was an attempt to make matrices module compatible with the domains module (`polys/domains` in sympy), but route appears to be abandoned (produced code even was marked as deprecated in sympy, see `sympy/matrices/densesolve.py` and so on).

jksuom:
> for polynomial ring over ZZ - the current form of rref() will not make any sense.

I agree. Probably Groebner bases should be used instead.

jksuom:
I think that it should be possible to generalize the Gröbner construction to polynomials over Z.

skirpichev:
@asmeurer, I will (hopefully) merge this work (a stalled forever PR in sympy) on terms of the sympy license.  Please reply, if you disagree.

skirpichev:
Maybe the cheapest alternative for `rref` change will be implementation `GF(p**n)`, i.e. for prime powers.  The tricky part is construction of an irreducible polynomial in `GF(p)[x]` of degree `q = p**n)`.  In our application `n=2**k`, maybe this does simplify things somehow?

jksuom:
The irreducible polynomial in `GF(p)[x]` that is needed for `GF(p**n)` is of degree `n`, not `q = p**n`. All such polynomials are factors of `x**(q-1) - 1` in `GF(p)[x]`. I would consider factoring that polynomial and choosing a factor of degree `n`.

skirpichev:
> of degree n, not q = p**n

Yes, of course.

>  All such polynomials are factors of `x**(q-1) - 1` in `GF(p)[x]` ...  consider factoring

Given a degree of that polynomial, I don't think is an efficient method.

I was thinking that in the particular case of `n = 2**k` there might be a simpler and more efficient way.  Something like `x**n - a` (as for `k=1`) or `x**n + x + 1`.

skirpichev:
Here some benchmarks of suggested generic method:
```
In [1]: R, x = ring('x', FF(2))                                                                                                        

In [2]: %time n = 2; q = R.domain.mod**n; p = x**(q-1) - 1; pprint([_[0] for _ in p.factor_list()[1] if _[0].degree() == n])           
[x**2 + x + 1 mod 2]
CPU times: user 12 ms, sys: 0 ns, total: 12 ms
Wall time: 36.8 ms

In [3]: %time n = 3; q = R.domain.mod**n; p = x**(q-1) - 1; pprint([_[0] for _ in p.factor_list()[1] if _[0].degree() == n])           
[x**3 + x + 1 mod 2, x**3 + x**2 + 1 mod 2]
CPU times: user 24 ms, sys: 8 ms, total: 32 ms
Wall time: 64 ms

In [4]: %time n = 4; q = R.domain.mod**n; p = x**(q-1) - 1; pprint([_[0] for _ in p.factor_list()[1] if _[0].degree() == n])           
[x**4 + x + 1 mod 2, x**4 + x**3 + 1 mod 2, x**4 + x**3 + x**2 + x + 1 mod 2]
CPU times: user 4.18 s, sys: 12 ms, total: 4.19 s
Wall time: 4.29 s
```
Already for `n=5` the algorithm is too slow to be practical.

jksuom:
For `n > 4`, I suspect that random trial polynomials would be better than polynomials with only few terms such as `x**n + x + 1`.

skirpichev:
> suspect that random trial polynomials would be better than polynomials with only few terms ...

Yes, that sounds as a plan.  Not sure, maybe there must be a variant for explicit construction of the field (i.e. a list of coefficients for irreducible polynomial `mod`).

BTW, while working on #761, I found that there are comparison operation defined for `ModularInteger`.  I believe, that's wrong.  Finite fields aren't ordered fields.  So, I think that these magic methods must be removed (like I did before for `AlgebraicElement`) and instead `is_positive/negative` methods may be reimplemented for the `FiniteField` class.  (Unfortunately, `is_positive/negative` - also somewhat misleading methods (see #14806.)

jksuom:
> a variant for explicit construction of the field

The construction of [Conway polynomials](https://en.wikipedia.org/wiki/Conway_polynomial_(finite_fields)) is explicit, in a way, but not very efficient. The idea is to test all polynomials of degree n in a fixed order and then choose the first irreducible one.

> these magic methods must be removed 

Yes, I don't think that they could be used for anything useful.

skirpichev:
> The construction of Conway polynomials is explicit, in a way, but not very efficient.

No.  I was thinking about more simple idea: allow for `FiniteField` constructor syntax like `FiniteField(p, ilist)`, where `ilist` - a list of `n+1` coefficients for an irreducible polynomial.  That might be useful to have a more deterministic way for creation of Galois fields.  At least for testing purposes.

skirpichev:
After #762 the Diofant has initial support for finite fields with non-prime orders.  If irreducible polynomial has not been provided - it will be generated by `gf_irreducible()`, but with fixed random seed. So, a kind of deterministic algorithm...  Unfortunately, this PR still doesn't work on top of that new code, without `rref()` modifications.

BTW, there no str/pretty printers for new finite field elements.  As we can assign nonnegative integer to any finite field element - may be just print this value, or there are better options?  The current `str` printer for `ModularInteger` prints things like `2 mod 3`, which is more suitable for finite rings.

jksuom:
I think that an element of `GF(p**n)` could be printed as a polynomial of degree < n in a generator (e.g. `alpha`) over `GF(p)`.

skirpichev:
> I think that an element of GF(p**n) could be printed as a polynomial of degree < n in a generator (e.g. alpha) over GF(p).

Yeah, you can see [how printed GF elements in the Mathematica](https://reference.wolfram.com/language/FiniteFields/tutorial/FiniteFields.html), so the correct output for `repr` would be something like `FF(3, [1, 0, 0, 1, 2])([1, 2, 1])`.  But this representation seems to be too verbose for `latex`/`pretty` and even `str` printers.

skirpichev:
Reopened in https://github.com/diofant/diofant/pull/876

PR #716, opened 2018-11-14
Using dist:xenial in Travis CI

skirpichev:
https://docs.travis-ci.com/user/reference/xenial/#python-support

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
   - pypy3.5
 env:
   global:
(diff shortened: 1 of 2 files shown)


skirpichev:
It seems, parsing module is broken on new minor releases of CPython.  At least, in 3.7.1 and 3.6.7.  Maybe [this tokenize change](https://docs.python.org/3/whatsnew/3.6.html#notable-changes-in-python-3-6-7) was the reason.  BTW, SymPy was affected too, so, @asmeurer, you may be interested (there is no issue opened so far).

PR #770, opened 2019-01-23
integerals: correct derivation() to return a poly over a field

skirpichev:
@jksuom, this is an alternative (single line patch) to https://github.com/sympy/sympy/pull/15826.

### DIFF
--- a/diofant/integrals/risch.py
+++ b/diofant/integrals/risch.py
@@ -790,7 +790,7 @@ def derivation(p, DE, coefficientD=False, basic=False):
     if basic:
         r = 0
     else:
-        r = Poly(0, DE.t)
+        r = Poly(0, DE.t, field=True)
 
     t = DE.t
     if coefficientD:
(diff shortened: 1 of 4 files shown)


diofant/integrals/risch.py
```diff
@@ -965,29 +965,29 @@ def hermite_reduce(a, d, DE):
...
+        ds_dms, _ = ds.div(dms)
         a = c.as_poly(DE.t) - (db*ds_dms).as_poly(DE.t)
 
         ga = ga*dm + b*gd
         gd = gd*dm
         ga, gd = ga.cancel(gd, include=True)
         dm = dm2
 
-    d = ds
+        d = ds
```
skirpichev: Maybe this was just a wrong indentation.

jksuom: This may take care of the case where the denominator is square-free, but I am not sure of the general case. I suspect that the issue could reappear in the computation of dm2. I would prefer a solution where the polynomials have coefficients in a field.

skirpichev:
I have found that for this issue it's enough to ensure `derivation()` return a `Poly` over a field.  I think, that with this fix issue will not reappear after entering the while loop in `hermite_reduce()`.

skirpichev:
Anyway, even if this doesn't solve all potential problems with `hermite_reduce()`, new fix is more consistent: it ensures that the whole function output has desired form.

PR #790, opened 2019-02-08
Misc fixes

skirpichev:


skirpichev:
I have found some discrepancies in definitions of gcd for dense and sparse representations, due to the [PolyElement._gcd_monom](https://github.com/diofant/diofant/blob/7c7bb68c8e3922c5d226c368a01d9aad5a5d1dec/diofant/polys/rings.py#L1969) method.  Take an example in SymPy (Out[2] seems to be incorrect):
```python
In [1]: R, x = ring('x', QQ)

In [2]: (3*x/2).gcd(R(QQ(9, 4)))
Out[2]: 3/4

In [3]: R.dmp_gcd(3*x/2, R(QQ(9, 4)))
Out[3]: 1
``` 

I think, this happens only for domains, which are fields (e.g. QQ).  I can't find an issue for sympy, @jksuom, do you think this is a bug?

d7d3bab951b is an attempt to fix, though it breaks some tests in the integrals module.

jksuom:
I agree. It is not mathematically wrong (any non-zero element of QQ is a gcd), but I think that it should be canonical in an algebraic library, hence 1 instead of 3/4.

skirpichev:
I'm not sure about broken test for `weak_normalizer()`.

jksuom:
It looks like a gcd canonicalization gone wrong.

skirpichev:
Are you sure the output is wrong, not just ugly?

jksuom:
I don't think it is mathematically wrong, but it would probably look better if the gcd's of polynomials were monic.

skirpichev:
Thank you for help.  I think, almost everything is ready to start switching Poly to using sparse representation.

jksuom:
You are welcome.

skirpichev:
I've found more differences for dense vs sparse representations.

1.   PolyElement._gcd_zero() vs _dmp_ff_trivial_gcd() (fixed in #795).  It seems, the first correctly reflects only _dmp_rr_trivial_gcd(), i.e. when the domain is a ring.  An example:
```
In [1]: R, x = ring('x', QQ.poly_ring(y))

In [2]: R(y).gcd(R(0))
Out[2]: y

In [3]: R.dmp_gcd(R(y), R(0))
Out[3]: y

In [4]: R, x = ring('x', QQ.frac_field(y))

In [5]: R(y).gcd(R(0))
Out[5]: y

In [6]: R.dmp_gcd(R(y), R(0))
Out[6]: 1
```

2. This case looks more severe: division for polynomials over a ring.
```
In [7]: R, x = ring('x', ZZ)

In [8]: f, g = 5*x**2 + 10*x + 3, 2*x + 2

In [9]: divmod(f, g)
Out[9]: (5, 5*x**2 - 7)

In [10]: R.dmp_div(f, g)
Out[10]: (0, 5*x**2 + 10*x + 3)
```

jksuom:
I think that the ring `R = ZZ[x]` is not Euclidean, so there is no canonical way of finding `q` and `r` that satisfy the equation `f = q*g + r`. The best one could expect is that, if `f` is divisible by `g` in `R`, then `q` is the quotient and `r = 0`. Otherwise, one can only say that the results are 'implementation dependent' and they should be depended on too much.

It is interesting that `R` is a unique factorization domain. Hence, `gcd` and `lcm` are defined for its elements. However, one cannot expect that `gcd` could be constructed by the Euclidean algorithm in `R`. (`QQ[x]`, which is Euclidean, shall be used instead.)

skirpichev:
On Thu, Feb 21, 2019 at 05:15:04AM -0800, Kalevi Suominen wrote:
>    I think that the ring R = ZZ[x] is not Euclidean, so there is no canonical
>    way of finding q and r that satisfy the equation f = q*g + r. The best one
>    could expect is that, if f is divisible by g in R, then q is the quotient
>    and r = 0.

>    Otherwise, one can only say that the results are
>    'implementation dependent' and they should be depended on too much.

Of course.  The problem is which variant is better.  Inconsistency is an
issue, from my point of view.  Fortunately, it doesn't break too much
tests, when I changed Poly to use sparse representation.

PR #795, opened 2019-02-17
Poly use sparse polynomial representation

skirpichev:
- [x] cleanup polytools tests (check dense representations?), merge separately
- [x] less conversions with ring.to/from_expr()
- [x] simplify the main commit (~~zero workarounds?~~)
- [x] better fixes for messy code (~~ode.py~~, risch.py)
- [x] div incompatibility, make consistent vs dense
- [x] compare performance on the test suite after this transition
- [x] release notes
- [x] redo documentation (L1 gone, OR?)


diofant/integrals/risch.py
```diff
@@ -762,7 +762,9 @@ def as_poly_1t(p, t, z):
         # issue sympy/sympy#4950
         raise NotImplementedError(e)
     # Compute the negative degree parts.
-    one_t_part = Poly.from_list(reversed(one_t_part.rep.rep), *one_t_part.gens,
```
skirpichev: @asmeurer, it seems this code rely on the bug of DMP.split(), which does preserve leading zeros (e.g. DMP.rep, coming from `split()`, may be like `[0, 2]` - such DMP object **is equal** to the DMP polynomial with rep `[2]`).  Is that really intentional thing?

asmeurer: I have no idea. This is a relevant issue though https://github.com/sympy/sympy/issues/5131

skirpichev: >  This is a relevant issue though sympy/sympy#5131

Yes, I'm aware of.  The problem is that this "hackish" solution is working due to the bug (in DMP.slice() or dmp_slice_in(), whatever).

diofant/solvers/ode.py
```diff
@@ -3792,16 +3792,12 @@ def _frobenius(n, m, p0, q0, p, q, x0, x, c, check=None):
...
-        if ser.is_polynomial(x) and Poly(ser, x).degree() <= n:
-            if x0:
-                ser = ser.subs({x: x + x0})
-            dict_ = Poly(ser, x).as_dict()
-        # Order term present
-        else:
+        if not ser.is_polynomial(x) or Poly(ser, x).degree() > n:
             tseries = series(ser, x=x, x0=x0, n=n + 1)
-            # Removing order
-            dict_ = Poly(list(ordered(tseries.args))[: -1], x).as_dict()
```
skirpichev: I also think, that this "removing" is a meaningless part of code.

This "works" in sympy only due to very low test coverage (one test, that I have found!).  In this example, the tseries is just a `4*x**4` and "removing order" give us just a `Poly(4, x)`.  Perhaps, there was assumed something like `Poly(tseries.removeO(), x)`?

skirpichev:
There are still several interfaces, which use dense representation, notably Poly.cancel.  But it shouldn't be too hard to port this function to use PolyElement.cancel.  That will solve sympy/sympy#16222.

skirpichev:
#813

PR #822, opened 2019-03-26
Optimize polynomial root isolation and refinement

skirpichev:


skirpichev:
The f834004 seems to be promising for big degrees.  For example, on this branch:
```python
In [2]: e = x**40 - 15*x**17 - 21*x**3 + 11
      : 
      : R = ZZ.poly_ring(x)
      : p = R.from_expr(e)
      : 

In [3]: %%time
      : complexes = R.dup_isolate_complex_roots_sqf(p, blackbox=True)
      : 
      : 
CPU times: user 2min 2s, sys: 180 ms, total: 2min 2s
Wall time: 2min 2s
```
While on the master:
```
In [4]: %%time
      : complexes = R.dup_isolate_complex_roots_sqf(p, blackbox=True)
      : 
      : 
CPU times: user 2min 52s, sys: 352 ms, total: 2min 52s
Wall time: 2min 52s
```

For low degrees - more testing required.  It looks, like transformation overhead kills its benefits here.

The missing part in this commit: scaling back ay and by, taking a square root.  For this application it should be possible if
  1) we take even number of horizontal bisections (this may require additional step)
  2) or if we inject sqrt(2) into the coefficient's domain.

PR #831, opened 2019-04-09
polys: workaround small coefficients, due to cancellation in inexact domains

skirpichev:
This was inspired by sympy/sympy#16596.

The idea was to replace test for being nonzero to something like `domain.almosteq(coeff, 0)`, if the `domain` of the coefficient `coeff` is inexact.  Of course, this principle must be applied in a more systematic way across all arithmetic stuff of the rings module.  Maybe the `__eq__` helper must be redefined for the case of polynomial rings over inexact domains (i.e. `a == 0` will mean all coeffs are almosteq to 0 for default tolerance)?

@jksuom, let me know what do you think about.  (BTW, your addition to the ~~FractionField~~FracField doesn't look right.  Perhaps, this should go to the `CompositeDomain` class or something.)


jksuom:
I think that something like `almosteq` is not easy to implement because the rounding error is relative to the leading coefficients of dividend and divisor. If they are small, then the error is even smaller. There is no absolute measure. Replacing `a - (a/b)*b` by zero should be a safe solution.

I admit that adding is_Exact to FractionField was a provisional (and minimal) solution to solve one of the issues. Your idea of adding it to CompositeDomain sounds good.

skirpichev:
On Tue, Apr 09, 2019 at 11:40:17PM -0700, Kalevi Suominen wrote:
>    I think that something like almosteq is not easy to implement because the
>    rounding error is relative to the leading coefficients of dividend and
>    divisor. If they are small, then the error is even smaller. There is no
>    absolute measure. Replacing a - (a/b)*b by zero should be a safe solution.

Hmm, thank you, seems reasonable.  So, the fix should be specific to the
div() helper.

I'm still unsure why it would be a safe solution.  Maybe the safest one
- forbid using inexact domains in algorithmic code of other modules,
like solvers or integrals (do conversion to an exact domain before such code
and the reversed one - after).  Most (all?) algorithms assume something more
rigorous about the coefficients domain (i.e. it's a ring, an integral
domain, a PID or something else specific).

>    I admit that adding is_Exact to FractionField was a provisional (and
>    minimal) solution to solve one of the issues. Your idea of adding it to
>    CompositeDomain sounds good.

Also, I don't sure PolynomialDivisionFailed exception in the dup_ff_div()
will be eventually triggered after your changes (maybe for EX domain?).
There are no coverage tests so far.

jksuom:
> the fix should be specific to the div() helper.

That is what I thought.

> why it would be a safe solution

I mean that it is a step in the right direction. It does not fix all the problems with rounding errors.

>  forbid using inexact domains in algorithmic code of other modules

This is probably not necessary. I think that a lot work has been done, for example, in linear algebra, to find reasonable solutions even in inexact domains. (Though I am actually not familiar with those matters.)

PR #857, opened 2019-06-28
Misc fixes

skirpichev:


skirpichev:
@jksuom, I've cherry-picked your fix for sympy/sympy#10996.  Also, this pr fixes problem in the modulargcd module, which, I'm sure, present in the sympy as well.

jksuom:
Thanks for looking into modulargcd. I only considered heuristicgcd. (It seems that no-one else understood the fix. Perhaps it looked too simple.)

PR #902, opened 2019-10-24
Misc fixes

skirpichev:


diofant/utilities/codegen.py
```diff
@@ -1369,7 +1369,8 @@ def _call_printer(self, routine):
                 raise CodeGenError("unexpected object in Routine results")
 
             constants, not_supported, oct_expr = octave_code(result.expr,
-                                                             assign_to=assign_to, human=False)
+                                                             assign_to=assign_to,
+                                                             human=False, inline=False)
```
skirpichev: @cbm755, probably that was a bug in OctaveCodeGen, which is still valid for SymPy.  With inline=True - constants are never generated.

cbm755: Thank you!  I'll try to get around to fixing this.

skirpichev:
@namannimmo10, sorry for late answer (I don't get notifications from the sympy repo): there is nothing to do in the diofant to fix issue https://github.com/sympy/sympy/issues/17792.  This pull request does all required.  I'm not sure if exactly same modifications are suitable for sympy, but probably so.

PR #914, opened 2019-11-12
Implement variable elimination by eliminate() function

skirpichev:
- [ ] polysys or polytools?
- [ ] add to the tutorial

### DIFF
--- a/diofant/solvers/__init__.py
+++ b/diofant/solvers/__init__.py
@@ -8,6 +8,6 @@ from .inequalities import reduce_inequalities
 from .ode import checkodesol, classify_ode, dsolve, homogeneous_order
 from .pde import (checkpdesol, classify_pde, pde_separate, pde_separate_add,
                   pde_separate_mul, pdsolve)
-from .polysys import solve_linear_system, solve_poly_system
+from .polysys import eliminate, solve_linear_system, solve_poly_system
 from .recurr import rsolve, rsolve_hyper, rsolve_poly, rsolve_ratio
 from .solvers import checksol, solve, solve_linear
--- a/diofant/solvers/polysys.py
+++ b/diofant/solvers/polysys.py
@@ -1,5 +1,6 @@
 """Solvers of systems of polynomial equations. """
 
+from ..core import Equality
 from ..domains import EX
 from ..matrices import Matrix
 from ..polys import groebner, poly, sring
@@ -10,7 +11,7 @@ from ..simplify import simplify
 from ..utilities import default_sort_key
 
 
-__all__ = 'solve_linear_system', 'solve_poly_system'
+__all__ = 'solve_linear_system', 'solve_poly_system', 'eliminate'
 
 
 def solve_linear_system(system, *symbols, **flags):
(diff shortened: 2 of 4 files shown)


skirpichev:
I'm little skeptical about this function.  This is a good example of poorly documented part of the Mathematica.  The [docs](https://reference.wolfram.com/language/ref/Eliminate.html) are misleading user to think that the solution of the obtained set of equations - match the original set of equations.

But this is not the case, as mentioned in the "Possible Issues" section, e.g.:
```
In [1]: eqns = [x*z - y, x**2 + x**2*z**2 - 1]

In [2]: eliminate(eqns, z)
Out[2]: 
⎡ 2    2    ⎤
⎣x  + y  - 1⎦

In [3]: solve(eqns)
Out[3]: 
⎡⎧          ________                     ⎫  ⎧         ________                
⎢⎪         ╱    2                        ⎪  ⎪        ╱    2                   
⎢⎪        ╱    z                         ⎪  ⎪       ╱    z                    
⎢⎪   -   ╱   ──────              ________⎪  ⎪      ╱   ──────            _____
⎢⎨      ╱     2                 ╱    2   ⎬  ⎨     ╱     2               ╱    2
⎢⎪    ╲╱     z  + 1            ╱    z    ⎪  ⎪   ╲╱     z  + 1          ╱    z 
⎢⎪x: ────────────────, y: -   ╱   ────── ⎪, ⎪x: ──────────────, y:    ╱   ────
⎢⎪          z                ╱     2     ⎪  ⎪         z              ╱     2  
⎣⎩                         ╲╱     z  + 1 ⎭  ⎩                      ╲╱     z  +

   ⎫                                         ⎤
   ⎪                                         ⎥
   ⎪                                         ⎥
___⎪                                         ⎥
   ⎬                                         ⎥
   ⎪                                         ⎥
── ⎪, {x: -1, y: 0, z: 0}, {x: 1, y: 0, z: 0}⎥
   ⎪                                         ⎥
 1 ⎭                                         ⎦
```
From ``Out[2]`` it looks, that ``{x: 0, y: 1}`` - part of the solution of ``eqns``, while it's not a case.

@jksuom, what do you think (given that there is a [similar issue](https://github.com/sympy/sympy/issues/2720) in the sympy?  The current approach could be generalized to include also rational and algebraic functions, but I doubt this should be included.  Maybe it would be better to wait a CAD-based analog of the [Resolve](https://reference.wolfram.com/language/ref/Resolve.html) method?

jksuom:
> it looks, that {x: 0, y: 1} - part of the solution of eqns, while it's not a case.

This is unavoidable. A solution of the full system will also satisfy the eliminated system but not vice versa, in general. In the language of algebraic geometry, this means that the projection of an affine algebraic set (i.e. the set of all solutions of a system of equations) is not an algebraic set. The eliminated system will define the smallest algebraic set containing the projected image (known as the Zariski closure of the set).

In order to get full correspondence, one has to use projective algebraic sets. So, in the above example, we should consider the projective 3-space with variables u, x, y, z, and homogeneous equations `x*z - u*y` and `u**2*x**2 + x**2*z**2 - u**4` (from which the original system is obtained by setting `u = 1`). This system has the solution `{u: 0, x: 0, y: 1}`.

skirpichev:
> This is unavoidable. A solution of the full system will also satisfy the eliminated system but not vice versa, in general.

That was exactly what I mean.

> The eliminated system will define the smallest algebraic set containing the projected image (known as the Zariski closure of the set).

How far this could be generalized, besides of the algebraic geometry?

jksuom:
I think that similar results hold for complex (or real) analytic sets, only proofs are harder.

skirpichev:
So, do you think - this does make sense for non-algebraic sets too?

PS: I doubt that a special algorithm is required here.  Maybe a simple ``solve()`` wrapper would be better, instead of duplicating it's logic more or less.  IIUC, we can just ignore zero-dimensional solutions (e.g. ``{x: 1, y: 0, z: 0}`` above), that's all.

jksuom:
> this does make sense for non-algebraic sets too

To some amount and under certain conditions ("proper" projection maps). However, analytic sets are much harder to handle than algebraic sets. (I would rather not even try...)

>  I doubt that a special algorithm is required here.

One has to be careful with Gröbner bases. The variables have to be properly ordered. 'lex' order is the one that has to be used for elimination, but that is not necessarily the alphabetical lexical order.

PR #947, opened 2020-01-14
_symbolic_factor_list() combine same multiplicity factors for method='sqf'

skirpichev:


skirpichev:
@jksuom, @akritas, you may be interested in this bugfix (valid for the SymPy too).  I would appreciate review.

N.b., the [SO problem](https://github.com/sympy/sympy/issues/8695#issuecomment-522278244) seems to be related to some other issue with the SymPy code, which was fixed in the Diofant before.  I would guess, it may be https://github.com/diofant/diofant/pull/726.

@jksuom, regarding [this your comment](https://github.com/sympy/sympy/issues/8695#issuecomment-570774164) - there are reasons to use a different "symbolic" algorithm (a wrapper, around suitable Poly's method) in sqf_list() function and friends: symbolic powers and huge exponents, like `sqf_list((x**2 + 2*x + 1)**20000000000)`.  I doubt if this has any practical value.

jksuom:
> there are reasons to use a different "symbolic" algorithm

Agreed. I'm now aware of that.

PR #964, opened 2020-02-17
Fix polynomial algorithms for FF of non-prime order

skirpichev:
- [x] check rest of factorization methods in the galoistools module
     - [x] dup_gf_ddf_zassenhaus
     - [x] dup_gf_edf_zassenhaus
     - [x] dup_gf_ddf_shoup
     - [x] dup_gf_edf_shoup
- [x] check irreducibility tests
     - [x] dup_gf_irred_p_rabin
     - [x] dup_gf_irred_p_ben_or
- [x] add references for functions over GF
- [x] add tests with prime power orders for _all_ functions
- [ ] a better variant for FiniteField.order property (it will clash with PolynomialRing): cardinality or size?
- [x] cleanup & rebase
- [ ] benchmarks?
- [x] release notes

### DIFF
--- a/diofant/polys/galoistools.py
+++ b/diofant/polys/galoistools.py
@@ -439,9 +439,9 @@ def dup_gf_berlekamp(f, K):
 
 def dup_gf_ddf_zassenhaus(f, K):
     """
-    Cantor-Zassenhaus: Deterministic Distinct Degree Factorization
+    Cantor-Zassenhaus: Deterministic Distinct Degree Factorization.
 
-    Given a monic square-free polynomial ``f`` in ``GF(p)[x]``, computes
+    Given a monic square-free polynomial ``f`` in ``GF(q)[x]``, computes
     partial distinct degree factorization ``f_1 ... f_d`` of ``f`` where
     ``deg(f_i) != deg(f_j)`` for ``i != j``. The result is returned as a
     list of pairs ``(f_i, e_i)`` where ``deg(f_i) > 0`` and ``e_i > 0``
@@ -463,15 +463,20 @@ def dup_gf_ddf_zassenhaus(f, K):
     References
     ==========
 
-    * :cite:`Gathen1999modern`
-    * :cite:`Geddes1992algorithms`
+    * :cite:`Gathen1999modern`, algorithm 14.3
+    * :cite:`Geddes1992algorithms`, algorithm 8.8
+
+    See Also
+    ========
+
+    dup_gf_edf_zassenhaus
 
     """
-    factors, p = [], K.characteristic
+    factors, q = [], K.order
     g, x = [[K.one, K.zero]]*2
 
     for i in range(1, dmp_degree_in(f, 0, 0)//2 + 1):
-        g = dup_gf_pow_mod(g, p, f, K)
+        g = dup_gf_pow_mod(g, q, f, K)
         h = dmp_gcd(f, dmp_sub(g, x, 0, K), 0, K)
 
         if not dmp_one_p(h, 0, K):
@@ -488,9 +493,9 @@ def dup_gf_ddf_zassenhaus(f, K):
 
 def dup_gf_edf_zassenhaus(f, n, K):
     """
-    Cantor-Zassenhaus: Probabilistic Equal Degree Factorization
+    Cantor-Zassenhaus: Probabilistic Equal Degree Factorization.
 
-    Given a monic square-free polynomial ``f`` in ``GF(p)[x]`` and
+    Given a monic square-free polynomial ``f`` in ``GF(q)[x]`` and
     an integer ``n``, such that ``n`` divides ``deg(f)``, returns all
     irreducible factors ``f_1,...,f_d`` of ``f``, each of degree ``n``.
     EDF procedure gives complete factorization over Galois fields.
@@ -506,8 +511,12 @@ def dup_gf_edf_zassenhaus(f, n, K):
     References
     ==========
 
-    * :cite:`Gathen1999modern`
-    * :cite:`Geddes1992algorithms`
+    * :cite:`Geddes1992algorithms`, algorithm 8.9
+
+    See Also
+    ========
+
+    dup_gf_ddf_zassenhaus
 
     """
     factors = [f]
@@ -515,8 +524,8 @@ def dup_gf_edf_zassenhaus(f, n, K):
     if dmp_degree_in(f, 0, 0) <= n:
         return factors
 
+    p, q = K.characteristic, K.order
     N = dmp_degree_in(f, 0, 0) // n
-    p = K.characteristic
 
     while len(factors) < N:
         r = dup_gf_random(2*n - 1, K)
@@ -525,10 +534,10 @@ def dup_gf_edf_zassenhaus(f, n, K):
             h = r
 
             for i in range(1, n):
-                r = dup_gf_pow_mod(r, p, f, K)
+                r = dup_gf_pow_mod(r, q, f, K)
                 h = dmp_add(h, r, 0, K)
         else:
-            h = dup_gf_pow_mod(r, (p**n - 1) // 2, f, K)
+            h = dup_gf_pow_mod(r, (q**n - 1) // 2, f, K)
             h = dmp_add_term(h, -K.one, 0, 0, K)
 
         g = dmp_gcd(f, h, 0, K)
@@ -542,16 +551,19 @@ def dup_gf_edf_zassenhaus(f, n, K):
 
 def dup_gf_ddf_shoup(f, K):
     """
-    Kaltofen-Shoup: Deterministic Distinct Degree Factorization
+    Kaltofen-Shoup: Deterministic Distinct Degree Factorization.
 
-    Given a monic square-free polynomial ``f`` in ``GF(p)[x]``, computes
-    partial distinct degree factorization ``f_1,...,f_d`` of ``f`` where
+    Given a monic square-free polynomial ``f`` in ``GF(q)[x]``, computes
+    partial distinct degree factorization ``f_1 ... f_d`` of ``f`` where
     ``deg(f_i) != deg(f_j)`` for ``i != j``. The result is returned as a
     list of pairs ``(f_i, e_i)`` where ``deg(f_i) > 0`` and ``e_i > 0``
     is an argument to the equal degree factorization routine.
 
+    Notes
+    =====
+
     This algorithm is an improved version of Zassenhaus algorithm for
-    large ``deg(f)`` and modulus ``p`` (especially for ``deg(f) ~ lg(p)``).
+    large ``deg(f)`` and order ``q`` (especially for ``deg(f) ~ lg(q)``).
 
     Examples
     ========
@@ -564,25 +576,30 @@ def dup_gf_ddf_shoup(f, K):
     References
     ==========
 
-    * :cite:`Kaltofen1998subquadratic`
+    * :cite:`Kaltofen1998subquadratic`, algorithm D
     * :cite:`Shoup1995factor`
     * :cite:`Gathen1992frobenious`
 
+    See Also
+    ========
+
+    dup_gf_edf_shoup
+
     """
-    n, p = dmp_degree_in(f, 0, 0), K.characteristic
+    n, q = dmp_degree_in(f, 0, 0), K.order
     k = math.ceil(math.sqrt(n//2))
     x = [K.one, K.zero]
 
-    h = dup_gf_pow_mod(x, p, f, K)
+    h = dup_gf_pow_mod(x, q, f, K)
 
-    # U[i] = x**(p**i)
+    # U[i] = x**(q**i)
     U = [x, h] + [K.zero]*(k - 1)
 
     for i in range(2, k + 1):
         U[i] = dup_gf_compose_mod(U[i - 1], h, f, K)
 
     h, U = U[k], U[:k]
-    # V[i] = x**(p**(k*(i+1)))
+    # V[i] = x**(q**(k*(i+1)))
     V = [h] + [K.zero]*(k - 1)
 
     for i in range(1, k):
@@ -618,15 +635,18 @@ def dup_gf_ddf_shoup(f, K):
 
 def dup_gf_edf_shoup(f, n, K):
     """
-    Gathen-Shoup: Probabilistic Equal Degree Factorization
+    Gathen-Shoup: Probabilistic Equal Degree Factorization.
 
-    Given a monic square-free polynomial ``f`` in ``GF(p)[x]`` and integer
-    ``n`` such that ``n`` divides ``deg(f)``, returns all irreducible factors
-    ``f_1,...,f_d`` of ``f``, each of degree ``n``. This is a complete
-    factorization over Galois fields.
+    Given a monic square-free polynomial ``f`` in ``GF(q)[x]`` and
+    an integer ``n``, such that ``n`` divides ``deg(f)``, returns all
+    irreducible factors ``f_1,...,f_d`` of ``f``, each of degree ``n``.
+    EDF procedure gives complete factorization over Galois fields.
+
+    Notes
+    =====
 
     This algorithm is an improved version of Zassenhaus algorithm for
-    large ``deg(f)`` and modulus ``p`` (especially for ``deg(f) ~ lg(p)``).
+    large ``deg(f)`` and order ``q`` (especially for ``deg(f) ~ lg(q)``).
 
     Examples
     ========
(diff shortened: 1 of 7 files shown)


skirpichev:
The problem with Galois fields, that integers are coerced to the GF elements per default, using [some bijection](https://github.com/diofant/diofant/blob/cfdfce09692a9437a1f93da5506a835488a4e8c3/diofant/domains/finitefield.py#L170) from {0,..,order-1} to GF elements.

Then, for example, in case of `F8 = GF(2, [1, 0, 1, 1])` we have `F8(1) = F8.one = GF(2, [1, 0, 1, 1])([1])` and `F8(2) = GF(2, [1, 0, 1, 1])([1, 0])`.  If we apply such coercion rules for `F8(1)*2`, we will have `F8(1)*F8(2) == F8(2)`.  But we could also could treat this case as a repeated addition: `F8(1)*2 == F8(1) + F8(1) = F8(0)`.

I suspect, most of PoiyElement's methods, which do arithmetics with raw ground domain elements - assume the second meaning.  E.g. [`_square()`](https://github.com/diofant/diofant/blob/cfdfce09692a9437a1f93da5506a835488a4e8c3/diofant/polys/rings.py#L1101) or [`diff()`](https://github.com/diofant/diofant/blob/cfdfce09692a9437a1f93da5506a835488a4e8c3/diofant/polys/rings.py#L1934).

The current pull request implements a special case for integer multiplication for the GaloisFieldElement (i.e. for x in GF: `x*integer` is an addition of `integer` clones `x`).  Maybe, this special case must be implemented in the PolyElement instead?

@jksuom, I would appreciate your opinion on this.  Maybe you will dislike the idea of using integer encoding at all, but I believe it's not the source of trouble...

@sylee957, it seems (from https://github.com/sympy/sympy/issues/18693) you are interested in implementing GF's of prime power orders for SymPy,  As you can see, this was implemented in the Diofant using quotient ring elements (just like AlgebraicElement's).  You should have everything you need in the current galoistools.py module of the SymPy (in fact, in the Diofant - most methods here were scrapped recently, so finite fields are not anymore a very special case for the dense recursive polynomial representation), i.e. there is a method to compute irreducible polynomial over prime-order GF, which can be used to represent the general `GF[p^n]` case. (However, I would appreciate a deterministic algorithm instead.)  BTW, you can see that this PR fix a trivial bug for the square-free factorization algorithm (non-zero characteristics).  Other special cases for factorization/irreducibility tests in the galoistools.py will require similar tiny fixes (I think, only the Shoup algorithm - will not work for non-prime orders).

jksuom:
I would define the product `FF(q) x ZZ -> FF(q)` as the combination

    FF(q) x ZZ -> FF(q) x FF(p) -> FF(q) x  FF(q) -> FF(q)

where `ZZ -> FF(p)` is the canonical homorphism (reduction mod p), `FF(p) -> FF(q)` is the canonical embedding, and the last map is the multiplication mapping of `FF(q)`. If I understand you correctly, this is your second method.

skirpichev:
On Thu, Feb 20, 2020 at 07:44:55AM -0800, Kalevi Suominen wrote:
>    If I understand you correctly, this is your second method.

Yes.  (Except that I don't reduce mod p.)  But maybe this belongs
to the PolyElement class rather the GaloisFieldElement?

jksuom:
> this belongs to the PolyElement class 

Yes, it looks like it.

skirpichev:
Hmm, I've experimented with using primitive polynomials to define GF.  The current random algorithm `dup_gf_irreducible()` with added test (with improvement, referenced in  sympy/sympy#18693) is too slow already for degree `n=5` in `FF(11)[x]`.  Maybe truly deterministic method will help, as Conway polynomials are primitive.

skirpichev:
> Yes, it looks like it.

Hmm, I'm less sure if that belongs to the PolyElement, in sense of arithmetics.

Let me try to explain the problem again.  In the Diofant, there is a one-to-one mapping from `{0, ..., q-1}` to `GF(q)` elements, where `q=p^m`.  So, before doing any arithmetics with ZZ elements - we make above conversion.

On another hand, sometimes we want to understand multiplication to ZZ, looking on `GF(q) x ZZ` as ZZ-algebra.  I don't think too much methods in the PolyElement class assume this meaning.  But take the `diff()`: the used formula `(x**n).diff(x)=n*x**(n-1)` is a direct consequence of the Leibniz rule and `x.diff(x)==1` (`n` is coming from the `n` terms of `x**(n-1)` arising).

So, one option is: fix these methods in PolyElement (maybe elsewhere in other functions, for this low-level polynomial representation):
```diff
diff --git a/diofant/polys/rings.py b/diofant/polys/rings.py
index a18e599cd3..a61b5fe9fa 100644
--- a/diofant/polys/rings.py
+++ b/diofant/polys/rings.py
@@ -1039,10 +1039,13 @@ def _pow_multinomial(self, n):
         terms = self.items()
         zero = self.ring.domain.zero
         poly = self.ring.zero
+        char = getattr(self.ring.domain, 'characteristic', 0)
 
         for multinomial, multinomial_coeff in multinomials:
             product_monom = zero_monom
             product_coeff = multinomial_coeff
+            if char:
+                product_coeff %= char
 
             for exp, (monom, coeff) in zip(multinomial, terms):
                 if exp:
@@ -1884,6 +1887,7 @@ def diff(self, x=0, m=1):
 
         """
         ring = self.ring
+        char = getattr(ring.domain, 'characteristic', 0)
         i = ring.index(x)
         x = ring.monomial_basis(i)
         x = x**m
@@ -1895,6 +1899,8 @@ def diff(self, x=0, m=1):
             if expv[i]:
                 e = expv/x
                 for j in range(expv[i], expv[i] - m, -1):
+                    if char:
+                        j %= char
                     coeff *= j
                 g[e] = coeff
         g._strip_zero()
```

@jksuom, are you assuming such fixes in the PolyElement?  BTW, probably the 'characteristic' should be just a Ring's class property (which in fact is a CommutativeRing).  We always can assume that derived classes can decide on this matter, right?

> I would define the product `FF(q) x ZZ -> FF(q)` as the combination
> `FF(q) x ZZ -> FF(q) x FF(p) -> FF(q) x  FF(q) -> FF(q)`

This is - another option.  Clearly, this belongs to the GaloisFieldElement, as it is right now.

Third option: forbid implicit `ZZ -> GF(q)` coercion with the mapping above.  Then we loose simple way of defining polynomials over `GF(q)` like `f=5*x**7 + 7*x + 2` for `GF(9)` and force users to be much more verbose.

jksuom:
It seems that integers could appear in two roles. In `5*x**7 + 7*x + 2`, they are used to *represent* a polynomial in `GF(9)[x]`. There is a representation mapping `[0, ..., q - 1] -> GF(q)` that is private to `GF(q)` and also a conversion `ZZ[x] -> GF(q)[x]` that makes use of the representation mapping. The second role would appear in the product context `GF(q) x ZZ -> GF(q)`. I think that here integers should be reduced mod p before multiplication so the map is biadditive. If that is not desirable, then the integers should first be explicitly converted to elements of `GF(q)`.

skirpichev:
Hmm, @jksuom, while porting `galoistools.py` code to prime power fields, I believe I have found a bug in the Berlekamp algorithm implementation.  Here is the wrong result from the sympy master session:
```
In [1]: R, x = ring('x', FF(2))                                                                                                        

In [2]: ans={}                                                                                                                         

In [3]: f = R.from_dense([1, 1, 1, 0, 1, 0, 1, 1, 0, 0, 1, 0, 0, 1, 0, 1, 0, 0, 1])                                                    

In [4]: f                                                                                                                              
Out[4]: x**18 + x**17 + x**16 + x**14 + x**12 + x**11 + x**8 + x**5 + x**3 + 1 mod 2

In [5]: from sympy.polys.polyconfig import using                                                                                       

In [6]:     for method in ('berlekamp', 'zassenhaus', 'shoup'): 
   ...:         with using(gf_factor_method=method): 
   ...:             ans[method] = f.factor_list() 
   ...:                                                                                                                                

In [7]: ans                                                                                                                            
Out[7]: 
{'berlekamp': (SymmetricModularIntegerMod2(1),
  [(x + 1 mod 2, 4),
   (x**14 + x**13 + x**12 + x**9 + x**7 + x**5 + x**4 + x**3 + 1 mod 2, 1)]),
 'zassenhaus': (SymmetricModularIntegerMod2(1),
  [(x + 1 mod 2, 4),
   (x**4 + x**3 + 1 mod 2, 1),
   (x**10 + x**8 + x**7 + x**5 + 1 mod 2, 1)]),
 'shoup': (SymmetricModularIntegerMod2(1),
  [(x + 1 mod 2, 4),
   (x**4 + x**3 + 1 mod 2, 1),
   (x**10 + x**8 + x**7 + x**5 + 1 mod 2, 1)])}
```

As you can see, the Berlekamp algorithm doesn't factor this reducible polynomial.  I can guess, that the bug is in the `dup_gf_Qbasis()` helper routine (`gf_Qbasis()` in sympy).

skirpichev:
Fixed by rewriting `dup_gf_Qbasis()` from scratch.

jksuom:
>  the bug is in the dup_gf_Qbasis()

That is my conclusion as well. It seems that the algorithm transforms the matrix whose nullspace is to be found to row-echelon form. Even columns are shuffled which makes it impossible to find the correct kernel vectors.

skirpichev:
Maybe there are less intrusive fixes, but don't see an easy way to correct the former code.

jksuom:
> don't see an easy way to correct the former code.

I think that there is something fundamentally wrong in that code.

skirpichev:
Maybe `rref/nullspace` from the matrices module should be used instead.  In the example above - `nullspace()` seems working, after adding support for `iszerofunc` kwarg.  The problem is - matrices module doesn't play well with element types.  E.g.:
```
In [1]: from diofant.polys.galoistools import dup_gf_frobenius_monomial_base, dup_gf_Qmatrix, dup_gf_Qbasis

In [2]: from diofant.polys.solvers import RawMatrix

In [3]: R, x = ring('x', FF(2))
      : f = R.from_dense([1, 1, 1, 0, 1, 0, 1, 1, 0, 0, 1, 0, 0, 1, 0, 1, 0, 0, 1])

In [4]: Q0 = dup_gf_Qmatrix(f.to_dense(), R.domain)
In [6]: Q1 = RawMatrix(Q0)

In [7]: Q2 = Q1 - eye(Q1.rows, cls=RawMatrix)

In [8]: Q2.nullspace(iszerofunc=lambda x: not x)
Out[8]: 
⎡⎡1⎤  ⎡0⎤  ⎡0⎤⎤
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢0⎥  ⎢1⎥  ⎢0⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢0⎥  ⎢1⎥  ⎢0⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢0⎥  ⎢0⎥  ⎢1⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢0⎥  ⎢1⎥  ⎢0⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢1⎥  ⎢0⎥  ⎢0⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢0⎥  ⎢0⎥  ⎢1⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢1⎥  ⎢0⎥  ⎢0⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢0⎥  ⎢1⎥  ⎢0⎥⎥
⎢⎢ ⎥, ⎢ ⎥, ⎢ ⎥⎥
⎢⎢0⎥  ⎢0⎥  ⎢1⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢1⎥  ⎢0⎥  ⎢0⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢1⎥  ⎢0⎥  ⎢0⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢0⎥  ⎢0⎥  ⎢1⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢1⎥  ⎢1⎥  ⎢1⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢1⎥  ⎢0⎥  ⎢0⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢1⎥  ⎢0⎥  ⎢0⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎢⎢0⎥  ⎢1⎥  ⎢0⎥⎥
⎢⎢ ⎥  ⎢ ⎥  ⎢ ⎥⎥
⎣⎣0⎦  ⎣0⎦  ⎣1⎦⎦
```

Maybe it doesn't matter, as 0's and 1's mean same for all domains.  Everything else (which is coming from `rref()`) seems to have a correct dtype for all elements.

But there are other tests, for which `rref()` seems to be broken.  I suspect, somewhere the overridden `_sympify` helper got ignored.

skirpichev:
Hmm, no.  Other tests also are working, you have to create unit matrix correctly.  E.g. with `eye(Q1.rows, cls=RawMatrix).applyfunc(R.domain.convert)`, c.f. above code.

I suspect, this is valid in the sympy too.

skirpichev:
Regarding the main issue with multiplication to integers: I think, that I should turn off the explicit bijection of integer ranges (up to the field order) to the GF elements during coercion.  I.e., for `F8=GF(8)`: F8(2) - maps to the 3-rd element of the F8 (in somewhat implementation-dependent order), but `F8.convert(2)` shouldn't do same.  Instead, the `F8.convert` should reduce `mod 2` before applying the above map (which will result in 0).  (The `F8.dtype.__int__()` magic may restore the int code of the element back, as it does now.)

The reason is simple.  `F8[x]` has also characteristic 2.  So, we should have `2*x = 0`, which will not happen if we will use integer coefficient polynomials to represent _all_ elements in `F8[x]`, i,e, if we apply `F8` element constructor during coercion `2*x == F8(2)*x`.

So, polynomial coefficients above the characteristic value should be converted to F8 explicitly, using type constructor (i.e. `F8(2)==F8.dtype(2)`).  I see no options.

skirpichev:
Another issue, related to SymPy, is [the wrong loop of the `gf_edf_zassenhaus()` for `p=2`](https://github.com/diofant/diofant/blob/0b3a17e9f0cffe633072bb6231fbd01168123cc1/diofant/polys/galoistools.py#L643-L650).  I think, this must be (for prime power `q=p**m`):
```python
         if p == 2:
             h = r
 
             for i in range(1, n):
                 r = dup_gf_pow_mod(r, q, f, K)
                 h = dmp_add(h, r, 0, K)
 
             g = dmp_gcd(f, h, 0, K)
```

Here is how to trigger the problem in SymPy:
```python
from sympy import FF, ring
from sympy.polys.polyconfig import using

F2 = FF(2)
R, x = ring('x', F2)

p = R.from_dense([1, 1, 1, 1, 1, 0, 0, 0, 1, 1, 0, 1, 1,
                  1, 1, 1, 1, 1, 0, 1, 1, 0, 0, 1, 1, 0, 1, 1, 0])

methods = 'shoup', 'zassenhaus'
ans = {}
for method in methods:
    with using(gf_factor_method=method):
        ans[method] = p.factor_list()

assert ans['shoup'] == ans['zassenhaus']
```
`gf_edf_zassenhaus()` got stuck in the loop, perhaps not forever.

Maybe it would be better to use the code snippet from `gf_edf_shoup()`, as the method seems to be same.

skirpichev:
I'm thinking on renaming of the FF.order property.  Currently, this will clash with the PolynomialRing.order.  I'm not sure about the future for the later property (maybe it must be removed), but a different name for FF's property wouldn't be bad in any case.  I see two options: size or cardinality.  Which would be better?  Any other options?

skirpichev:
Ok, thank you for some help with review.

skirpichev:
@jksuom, I think that mentioned above `gf_edf_zassenhaus()` was responsible for the sympy/sympy#16620.

jksuom:
Thanks for the reminder. I'll have to keep that in mind and look into it.

PR #976, opened 2020-03-15
Drop dup_gf_Qbasis() helper, reuse Matrix.nullspace() code

skirpichev:
This is an alternative fix to the problem with dup_gf_Qbasis(),
mentioned in the diofant/diofant#964.

Closes sympy/sympy#18749

### DIFF
--- a/diofant/polys/galoistools.py
+++ b/diofant/polys/galoistools.py
@@ -6,7 +6,7 @@ import random
 from ..ntheory import factorint
 from .densearith import (dmp_add, dmp_add_term, dmp_mul, dmp_quo, dmp_rem,
                          dmp_sqr, dmp_sub)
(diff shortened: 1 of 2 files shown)


skirpichev:
@jksuom, it seems the Berlekamp algorithm issue may be solved, using Matrix.nullspace() too.  This should work in the sympy, if you (sympy devs) don't break anything here since v0.7.6.

jksuom:
It seems that SymPy's `nullspace` and Knuth's algorithm `N` are essentially the same except that the latter works with transposed matrix and row vectors. It may also be slightly more efficient as it does not try to put the matrix into an echelon form while `nullspace` calls `rref`.

skirpichev:
Yes, maybe I should replace `nullspace()` with the new code.  But anyway, the code redundancy should be eliminated.

PR #977, opened 2020-03-16
Cleanup Domain interfaces

skirpichev:
- [x] release notes
- [x] cleanup
- [x] expand docstring, ~~make abstract method (like characteristic property)?~~

### DIFF
--- a/diofant/domains/algebraicfield.py
+++ b/diofant/domains/algebraicfield.py
@@ -22,7 +22,7 @@ _algebraic_numbers_cache = {}
 class AlgebraicField(CharacteristicZero, SimpleDomain, Field):
     """A class for representing algebraic number fields."""
 
-    is_AlgebraicField = is_Algebraic = True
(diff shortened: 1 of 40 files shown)


skirpichev:
I think, I have found the source of is_positive/is_negative/etc methods for Domain-derived classes.  This is used only for making gcd output - unique.  I believe, something like is_nonnegative has actual meaning of being unit normal and all other predicates must be removed per sympy/sympy#14806.

@jksuom, I would appreciate your opinion.  This PR introduce is_normal() method for the Ring class (which in the Diofant is actually - an integral domain).  Maybe the name should be more verbose (e.g. is_unit_normal) or even the interface could be private.

jksuom:
I fully agree with you. I have always thought that is_positive/is_negative is a bad and confusing choice for elements of general domains. As to the name, I think that is_normal is passable. I cannot think of a better name.

PR #1033, opened 2020-06-28
Protect PolyElement's from being modified

skirpichev:
- [x] remove `__hash__()` comment (more?)
- [ ] adapt more methods?
- [x] docs for PolyElement
- [x] release notes


skirpichev:
PolyElement's mutability poses serious problem for SymPy as well.  So far, as I see, it's unsolved yet (e.g. [this](https://github.com/sympy/sympy/blob/1a84db6a20ee606c0e6388267151f8d73145fa1c/sympy/polys/rings.py#L568-L572)).

@asmeurer, what do you think of this idea?  Overhead is as low as I can imagine.  Of course, `dict.update()` must be overloaded as well and so on (if PolyElement will subclass `dict` as now).

PR #1067, opened 2020-09-26
Solve inequalities in extended reals

skirpichev:


skirpichev:
Few tests seems to be wrong:
```
In [2]: reduce_inequalities(-1 + 1/abs(1/x - 1) < 0)
Out[2]: (0 < x ∧ x < 1/2) ∨ x < 0

In [3]: limit(-1 + 1/abs(1/x - 1) < 0, x, -oo)
Out[3]: false

In [4]: reduce_inequalities(x/(x + 1) > 1, x)
Out[4]: x < -1

In [5]: limit(x/(x + 1) > 1, x, -oo)
Out[5]: false
```

Probably, we should test infinities in the rational functions case.

PR #1076, opened 2020-10-19
Correct AlgebraicField.from_expr()

skirpichev:


skirpichev:
@jksuom, it seems you completely removed groebner-based algorithm of computing the primitive element in sympy/sympy#20274.  Did you try to make some benchmarks?  Is the groebner-based version always suboptimal, maybe there should be an option to select an algorithm, like for the minimal_polynomial()?

jksuom:
I did that because it looks like there is no guarantee that the groebner-based algorithm will work, not because the other algorithm would be more efficient (though I would expect so).

skirpichev:
> there is no guarantee that the groebner-based algorithm will work

There is, if it's implemented correctly (like the diofant's case, I hope).See [refs](https://diofant.readthedocs.io/en/latest/modules/polys.html#diofant.polys.numberfields.primitive_element), this is the algorithm 4.3 in [YN89](https://diofant.readthedocs.io/en/latest/sources.html#yokoyama1989primitive).

> though I would expect so

For me, it seems too.  But I would expect, that it depends on the algorithm for the groebner bases computation, as the [YN89](https://diofant.readthedocs.io/en/latest/sources.html#yokoyama1989primitive) said.

PR #1115, opened 2021-01-27
Misc fixes

skirpichev:


skirpichev:
@AaronStiff, I'm glad you are using part of this work, related to sympy/sympy#20902, but please remember that such sorting (either in the `solve_poly_inequality()` or `Poly.real_roots()`) is just a workaround.

I think that the correct solution - is fixing [is_disjoint()](https://github.com/diofant/diofant/blob/8a72d663035f1c27609a965d90217fbe65c1c44d/diofant/polys/rootisolation.py#L860-L862) method of RealInterval by using strict inequalities.

skirpichev:
@AaronStiff, please don't try to ping me in the sympy repo.  I will be not notified about, in general.

This time I've seen your [question](https://github.com/sympy/sympy/pull/20906#issuecomment-773985552), so I'll try to addess this.  The problem is that `solve_poly_inequalities()` assume that (distinct) real roots are sorted by value.  And such sorting should be expected after `RooOf._reals_sorted()`.  Unfortunately, it doesn't work if compared intervals intersect by boundaries, e.g. corresponding left and right ends are equal.  IIRC, in sympy/sympy#20902 the problem polynomial is something like `(x + 1)**3*(3*x + 1)`.  In this case, for distinct roots, we have intervals like `r1=[-1, -1]` (this root is from the ground domain, i.e. ZZ) and something like `r2=[-1, 0]` for `x=-1/3`.  Second root is not going to be improved by sorting helper, since `r1.b==r2.a`.

I think it's perfectly fine to keep improving roots in such cases, using strict inequalities in the `RealInterval.is_disjoint()`, as it was suggested.   I would appreciate @jksuom feedback (see above commit), for better safety.

AaronStiff:
Oh sorry about that. Thanks for explaining!

skirpichev:
Ok, rebased fix for RealInterval goes to https://github.com/diofant/diofant/pull/1117, merged.

PR #1128, opened 2021-03-12
Enable testing on CPython 3.10

### LINKED ISSUE #1119: Clear workarounds for nedbat/coveragepy#198
... if this will be fixed after 3.10.  An example: [this](https://github.com/diofant/diofant/blob/7f24859c9c0da37b3d060f212be66e7940bcb05a/diofant/core/add.py#L85).

Meanwhile we might consider a better workaround, which can be removed with optimizations like assert statement.

skirpichev:
Trying to fix #1119.  Coverage tests will run on 3.10.


skirpichev:
Seems to be working, but I'll wait for 3.10 wheels of "exports" requirements...

skirpichev:
Hmm, maybe codecov bug?  It seems codecov reports much more coverage regressions than the build artifact.

PR #1132, opened 2021-03-19
polys: square-free factorization for multivariate polynomials over GF

skirpichev:
- [x] ~~`.root(p)` helper for Monomial for `p`-th root?~~
- [x] ~~references, better than [this](https://raw.githubusercontent.com/PoslavskySV/rings/develop/rings/src/main/java/cc/redberry/rings/poly/multivar/MultivariateSquareFreeFactorization.java)?~~
- [x] move GF's examples to the public `sqf_list()` docstring?
- [x] more tests

### DIFF
--- a/diofant/polys/sqfreetools.py
+++ b/diofant/polys/sqfreetools.py
@@ -18,19 +18,20 @@ class _SQF:
         >>> R.sqf_list(x**5 + 2*x**4*y + x**3*y**2)
         (1, [(x + y, 2), (x, 3)])
 
-        >>> _, x = ring('x', FF(11))
-        >>> f = x**11 + 1
+        >>> R, x, y = ring('x y', FF(5))
 
-        Note that:
+        >>> f = x**5*y**5 + 1
 
-        >>> f.diff()
-        0 mod 11
+        Note that e.g.:
+
+        >>> f.diff(x)
+        0 mod 5
 
         This phenomenon doesn't happen in characteristic zero. However we can
         still compute square-free decomposition of ``f``:
 
-        >>> f.sqf_list()
-        (1 mod 11, [(x + 1 mod 11, 11)])
+        >>> R.sqf_list(f)
+        (1 mod 5, [(x*y + 1 mod 5, 5)])
 
         """
         domain = self.domain
@@ -46,31 +47,34 @@ class _SQF:
             return coeff, self._rr_yun0_sqf_list(f)
 
     def _gf_sqf_list(self, f):
-        """Compute square-free decomposition of the monic ``f`` in ``GF(q)[X]``."""
-        if self.is_multivariate:
-            raise NotImplementedError('multivariate polynomials over finite fields')
-        else:
-            return self._gf_musser_sqf_list(f)
+        """
+        Compute square-free decomposition of the monic ``f`` in ``GF(q)[X]``.
 
-    def _gf_musser_sqf_list(self, f):
-        """Compute square-free decomposition of the monic ``f`` in ``GF(q)[x]``.
+        Notes
+        =====
+
+        Uses a modified version of Musser's algorithm for square-free
+        decomposition of univariate polynomials over finite fields.
 
         References
         ==========
 
         * :cite:`Geddes1992algorithms`, algorithm 8.3
+        * :cite:`Musser1971factor`, p.54, algorithm V (modified)
 
         """
         domain = self.domain
 
-        n, factors, p = 1, [], domain.characteristic
+        n, factors, p = 1, [], int(domain.characteristic)
         m = int(domain.order // p)
 
         while not f.is_ground:
-            df = f.diff()
+            df = [f.diff(x) for x in self.gens]
 
-            if not df.is_zero:
-                g = self.gcd(f, df)
+            if not all(_.is_zero for _ in df):
+                g = f
+                for q in df:
+                    g = self.gcd(g, q)
                 h, f, i = f // g, g, 1
 
                 while not h.is_one:
@@ -87,10 +91,8 @@ class _SQF:
             n *= p
 
             g = self.zero
-            for i in range(f.degree()//p + 1):
-                k = (i*p,)
-                if k in f:
-                    g[(i,)] = f[k]**m
+            for monom, coeff in f.items():
+                g[(_ // p for _ in monom)] = coeff**m
             f = g
 
         return factors
--- a/diofant/tests/polys/test_sqfreetools.py
+++ b/diofant/tests/polys/test_sqfreetools.py
@@ -171,6 +171,42 @@ def test_dmp_sqf():
     assert (x**2 - 2).sqf_norm() == (1, x**2 - 2*sqrt(3)*x + 1,
                                      (x**4 - 10*x**2 + 1).set_domain(QQ))
 
+    R, x, y = ring('x y', FF(2))
+
+    f = y**2 + 1
+
+    assert f.is_squarefree is False
+    assert f.sqf_part() == y + 1
+    assert f.sqf_list() == (1, [(y + 1, 2)])
+
+    f = x**3 + x*y**2
+
+    assert f.is_squarefree is False
+    assert f.sqf_part() == x*(x + y)
+    assert f.sqf_list() == (1, [(x, 1), (x + y, 2)])
+
+    f1 = 1 + x**6*y**14 + x**2*y**4 + x**7
+    f2 = 1 + x**3*y**4 + x + y
+
+    assert f1.is_squarefree
+    assert f2.is_squarefree
+    assert (f1**2*f2**2).sqf_list() == (1, [(f1*f2, 2)])
+
+    R, x, y = ring('x y', FF(5))
+
+    f = x**5*y**5 + 1
+
+    assert f.is_squarefree is False
+    assert f.sqf_list() == (1, [(x*y + 1, 5)])
+
+    F8 = FF(8)
+    R, x, y = ring('x y', F8)
+
+    f = x**8*y**8 + 1
+
+    assert f.is_squarefree is False
+    assert f.sqf_list() == (1, [(x*y + 1, 8)])
+
     R, x, y = ring('x y', ZZ)
 
     assert R(0).sqf_part() == 0
@@ -211,11 +247,6 @@ def test_dmp_sqf():
 
     pytest.raises(DomainError, lambda: (x**2 + y**2).sqf_norm())
 
-    R, x, y = ring('x y', FF(2))
-
-    pytest.raises(NotImplementedError, lambda: (y**2 + 1).sqf_list())
-    pytest.raises(NotImplementedError, lambda: (x**3 + 2*x**2*y + x*y**2).sqf_part())
-
     R, x, y = ring('x y', QQ.algebraic_field(I))
 
     assert (x**2 + 2*I*x - 1).sqf_list() == (1, [(x + I, 2)])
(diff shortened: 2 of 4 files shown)


skirpichev:
@cathreya, that might partially fit your needs for multivariate polynomials over GF.  There are also better (Yun-based) algorithm, implemented in the L.Bernardin's "Factorization of Multivariate Polynomials over Finite Fields" dissertation.

skirpichev:
@jksuom, I think, there is a bug in SymPy.  I don't have a minimal example yet, but this hangs:
```
In [1]: D = FF(17)
In [2]: R = D.inject(x, y, z)
In [3]: a = R.from_sympy("11 + x^7*y^14 + z^7*x^14 + y^7")
In [4]: b = R.from_sympy("11 + 3*y^7*z^14 + 4*x^7*y^14 + 5*z^7")
In [5]: c = R.from_sympy("z*y^2*x^2 - 2*y^3*x - 1234*z^7*x^12*y^13")
In [6]: p = a**2*b**2*c**2
In [7]: pd = p.gcd(p.diff(0))
^C--- # after ~1hour
....
```
I suspect a problem somewhere in the arithmetic stuff (quotient/remainder), because if I try simple euclidean gcd (like [here](https://github.com/diofant/diofant/blob/0f922a78980f3644ba89bcc16e8c8108d94f7188/diofant/polys/univar.py#L182-L188)) - this still hands...

jksuom:
Multivariate polynomial rings are not Euclidean domains (so there is no sensible quo/rem algorithm for computing gcd). It seems that the implemented gcd algorithm is not valid for such polynomials. The relevant code in rings.py looks complicated and I have never tried to understand its working. I think that the correct way of computing the gcd of multivariate polynomials is to consider one variable at a time. (Make them univariate with respect to one of the variables with coefficients in the ring generated by the remaining variables.)

skirpichev:
On Sat, Mar 20, 2021 at 06:01:36AM -0700, Kalevi Suominen wrote:
>    Multivariate polynomial rings are not Euclidean domains (so there is no
>    sensible quo/rem algorithm for computing gcd).

Indeed, I forgot to make euclidead algorithm recursive (i.e. eject
all but one variables and so on).

>    It seems that the implemented gcd algorithm is not valid for such polynomials.

I doubt.  AFAIK, SymPy (like Diofant) uses Brown's PRS algorithm
to compute subresultants and gcd per default (ZZ/QQ and algebraic
number domains have specialized methods).

>    I think that the correct way of computing the gcd of
>    multivariate polynomials is to consider one variable at a time. (Make them
>    univariate with respect to one of the variables with coefficients in the
>    ring generated by the remaining variables.)

Yes, this work in this way.  Perhaps, it's more clear for the Diofant:
relevant helpers are _rr_prs_gcd() and _ff_prs_gcd() in euclidtools.py.

jksuom:
> gcd algorithm is not valid for such polynomials.

On closer look, it seems that the algorithm is correct in principle, though there may be something wrong in the implementation. I'll try to take some more time to investigate.

skirpichev:
On Sat, Mar 20, 2021 at 09:55:12AM -0700, Kalevi Suominen wrote:
>    On closer look, it seems that the algorithm is correct in principle,
>    though there may be something wrong in the implementation.

Maybe it just too slow here, not hangs?  Diofant's variant take more
memory than SymPy.  I'll try to run SymPy's version tonight.

Huge memory overhead may be eliminated, I hope, if we will save
only last subresultants, not all.  Only few last terms are required
to compute gcd and resultant.

>    I'll try to take some more time to investigate.

I'll try to reproduce this on a smaller example.  But so far - all I
tried looks correct, just slow.

jksuom:
> Maybe it just too slow here, not hangs?

That is what it now looks like to me. I tried to run the example `p.gcd(p.diff(0))` but quit after 28 hours. Then I took a simpler case that finished:
```
In [41]: r = a**2*b**2                                                         

In [42]: %time r.gcd(r.diff(0))                                                
CPU times: user 11h 39min 1s, sys: 14.1 s, total: 11h 39min 15s
Wall time: 11h 39min 17s
Out[42]: 9 mod 17*y**14*z**7*x**21 + (9 mod 17*y**28 + 11 mod 17*y**7*z**21 + 7 mod 17*z**14 + 12 mod 17*z**7)*x**14 + (11 mod 17*y**21*z**14 + 9 mod 17*y**21 + 7 mod 17*y**14*z**7 + 9 mod 17*y**14)*x**7 + 11 mod 17*y**14*z**14 + 2 mod 17*y**7*z**14 + 7 mod 17*y**7*z**7 + 12 mod 17*y**7 + 9 mod 17*z**7 + 13 mod 17
```
I didn't check the output but I would be surprised if it were incorrect.

skirpichev:
On Tue, Mar 23, 2021 at 02:08:27AM -0700, Kalevi Suominen wrote:
>      Maybe it just too slow here, not hangs?
>    That is what it now looks like to me. I tried to run the example
>    p.gcd(p.diff(0)) but quit after 28 hours.

Yes, I'm giving up as well.  I tried to run this for several days ~since
the first report.  Either it eat all dedicated memory (1Gb) or just
hangs (still running now).  That's possible with a patch:
diff --git a/sympy/polys/euclidtools.py b/sympy/polys/euclidtools.py
index c0b6d13e79..74e11b60a7 100644
--- a/sympy/polys/euclidtools.py
+++ b/sympy/polys/euclidtools.py
@@ -364,7 +364,7 @@ def dup_inner_subresultants(f, g, K):
 
     while h:
         k = dup_degree(h)
-        R.append(h)
+        R = [R[-1], h]
 
         f, g, m, d = g, h, k, m - k
 
@@ -381,7 +381,7 @@ def dup_inner_subresultants(f, g, K):
         else:
             c = -lc
 
-        S.append(-c)
+        S = [S[-1], -c]
 
     return R, S
 
@@ -503,7 +503,7 @@ def dmp_inner_subresultants(f, g, u, K):
 
     while not dmp_zero_p(h, u):
         k = dmp_degree(h, u)
-        R.append(h)
+        R = [R[-1], h]
 
         f, g, m, d = g, h, k, m - k
 
@@ -522,7 +522,7 @@ def dmp_inner_subresultants(f, g, u, K):
         else:
             c = dmp_neg(lc, v, K)
 
-        S.append(dmp_neg(c, v, K))
+        S = [S[-1], dmp_neg(c, v, K)]
 
     return R, S


>    Then I took a simpler case that finished:
> 
>  In [41]: r = a**2*b**2
> 
>  In [42]: %time r.gcd(r.diff(0))
>  CPU times: user 11h 39min 1s, sys: 14.1 s, total: 11h 39min 15s
>  Wall time: 11h 39min 17s
>  Out[42]: 9 mod 17*y**14*z**7*x**21 + (9 mod 17*y**28 + 11 mod 17*y**7*z**21 + 7 mod 17*z**14 + 12 mod 17*z**7)*x**14 + (11 mod 17*y**21*z**14 + 9 mod 17*y**21 + 7 mod 17*y**14*z**7 + 9 mod 17*y**14)*x**7 + 11 mod 17*y**14*z**14 + 2 mod 17*y**7*z**14 + 7 mod 17*y**7*z**7 + 12 mod 17*y**7 + 9 mod 17*z**7 + 13 mod 17
> 
>    I didn't check the output but I would be surprised if it were incorrect.

Yes, I've tried several such cases (with checks, I'll test this too) and I've seen
no errors.  Perhaps, algorithm just not optimal here and we
should use somthing like "R E Zippel. Probabilistic algorithms for sparse
polynomials. In Proceedings of the International Symposiumon on Symbolic
and Algebraic Computation, EUROSAM ‘79, pages 216–226, London, UK,
UK, 1979. Springer-Verlag."

Sorry for a false alert.

skirpichev:
> r = a**2*b**2                                                         
> In [42]: %time r.gcd(r.diff(0))   
>  I didn't check the output but I would be surprised if it were incorrect.

Hmm, perhaps I can surprise you...  Here is the answer in the Mathematica session (Wolfram Language 12.2.0 Engine):
```
In[12]:= a = -x^42*y^28*z^14 - 2*x^35*y^42*z^7 + 7*x^35*y^21*z^28 + 6*x^35*y^14*z^21 + 3*x^35*y^14*z^14 - x^28*y^56 - 3*x^28*y^35*z^21 - 2*x^28*y^35*z^7 - 5*x
^28*y^28*z^14 + x^28*y^28*z^7 - 8*x^28*y^14*z^42 - 4*x^28*y^7*z^35 - 2*x^28*y^7*z^28 + 8*x^28*z^28 + 8*x^28*z^21 + 2*x^28*z^14 + 7*x^21*y^49*z^14 - 2*x^21*y^4
9 + 6*x^21*y^42*z^7 - 2*x^21*y^42 + x^21*y^28*z^35 - 3*x^21*y^28*z^21 - 8*x^21*y^21*z^28 - 3*x^21*y^21*z^21 - 5*x^21*y^21*z^14 + 6*x^21*y^21*z^7 - x^21*y^14*z
^21 - 5*x^21*y^14*z^14 + 2*x^21*y^14*z^7 - 8*x^14*y^42*z^28 - 3*x^14*y^42*z^14 - x^14*y^42 - 4*x^14*y^35*z^21 - x^14*y^35*z^14 - 5*x^14*y^35*z^7 + x^14*y^35 +
 8*x^14*y^28*z^14 + 4*x^14*y^28*z^7 - 2*x^14*y^28 + x^14*y^21*z^35 - 6*x^14*y^14*z^35 - 8*x^14*y^14*z^28 - 4*x^14*y^14*z^21 - 3*x^14*y^7*z^28 + 6*x^14*y^7*z^2
1 - x^14*y^7*z^14 + 4*x^14*y^7*z^7 + 6*x^14*z^21 + 6*x^14*z^14 - 7*x^14*z^7 + x^7*y^35*z^28 + 7*x^7*y^35*z^14 - 6*x^7*y^28*z^28 - 8*x^7*y^28*z^21 - 3*x^7*y^28
*z^14 + 6*x^7*y^28*z^7 + 3*x^7*y^28 - 3*x^7*y^21*z^21 + 3*x^7*y^21*z^14 - 5*x^7*y^21*z^7 + 2*x^7*y^21 + 6*x^7*y^14*z^14 + x^7*y^14*z^7 - x^7*y^14 - 8*y^28*z^2
8 - 6*y^21*z^28 - 4*y^21*z^21 - 2*y^21*z^14 + y^14*z^28 - 3*y^14*z^21 - 2*y^14*z^14 + 8*y^14*z^7 + 2*y^14 - 8*y^7*z^21 + 2*y^7*z^14 + 6*y^7*z^7 - 7*y^7 - z^14
 - z^7 + 4;

In[13]:= PolynomialGCD[a, D[a, x], Modulus->17]

                7    7  14    7  21    14  28    7      14  7       7  7       7  14  7    21  14  7       14  14      7  14      14  14      7  21  14
Out[13]= 9 + 7 y  + x  y   + x  y   + x   y   + z  + 7 x   z  + 14 y  z  + 14 x  y   z  + x   y   z  + 14 x   z   + 4 y  z   + 5 y   z   + 5 x  y   z   +

        14  7  21
>    5 x   y  z
In[14]:= %13/.{x->0, y->0, z->0}                                                                                                                              

Out[14]= 9
```

skirpichev:
Hmm, here is what I get on the Diofant:
```
$ time python a.py
x**7*y**3*z**2 + x**5*y**10 + 5 mod 17*x**5*y**2*z**6 + 14 mod 17*x**5*z**4 +
7 mod 17*x**5*z**2 + 5 mod 17*x**3*y**9*z**4 + 14 mod 17*x**3*y**7*z**2 +
7 mod 17*x**3*y**7 + x**2*y**6 + 11 mod 17*x**2*y**3 + 5 mod 17*y**5*z**4 +
14 mod 17*y**3*z**2 + 7 mod 17*y**3 + 4 mod 17*y**2*z**4 + z**2 + 9 mod 17
real    183m55,396s
user    177m22,404s
sys     0m17,849s
```
for
```
$ cat a.py
from diofant import *

R, x, y, z = ring('x y z', FF(17))

a = x**5*z**2 + x**3*y**7 + y**3 + 11
b = 4*x**2*y**3 + 3*y**2*z**4 + 5*z**2 + 11
c = 7*x**2*y**3*z**7 + x**2*y*z + 15*x*y**3

e = a**2*b**2

r = e.gcd(e.diff(0))

print(sstr(r))
```
Doesn't match the Mathematica's output as well.  But if we divide on `5` - it does.  And, to my surprise, this computation was a little faster (~3 vs 11 hours) for old 2-core Intel Celeron @ 2.16GHz.

It doesn't look like Mathematica's output and SymPy's differ up to a constant.  So, I suspect there **can** be a problem.

skirpichev:
> I didn't check the output but I would be surprised if it were incorrect.

I've checked your answer with the Mathematica's PolynomialGCD.  It differ on a factor `1/9`.

jksuom:
> It differ on a factor 1/9

That is what I thought. I had many 9's where Mathematica had 1's.

skirpichev:
Mathematica's answer seems to be monic wrt lex order (as well as in the Diofant).  SymPy's - not.

jksuom:
Actually, I tried to simplify the computation by converting the polynomials to elements of `GF(17)[y, z}[x]`. Apparently, the result will not be monic when the coefficient domain is not a field.

Issue #1028, opened 2020-06-14
Consider page listing differences from Sympy?


matthew-brett:
I occaisionally come back to diofant to see how y'all are doing, but I really miss some description of how diofant currently differs from Sympy, why I might chose it instad of Sympy.   Would you consider making such a page - maybe for the docs?

skirpichev:
Thank you for your attention to the project.  Unfortunately, I have doubts on this feature.

1) I'm not sure why the SymPy is a special case.  There are no plans to support compatibility with the SymPy or to merge projects.  Should I document also differences with the Maple or the Maxima?
2) I don't follow the SymPy development and hardly able to do comparison.
3) The Diofant has no stable API and the SymPy also make changes fast - it won't be easy to keep suggested docs in sync.  Huge maintenance burden.

I'm trying to document changes (incl. compatibility breaks) in the release notes.  Some important (on my view) changes also are highlighted in release announcements on the mail list. And there are the API docs (the "Reference" section), mostly in a horrible state, yes.  Not sure how the suggested comparison fits this picture.

Maybe your suggestion does make sense as a blog post or something (where comparison will be restricted to some versions), I don't know.  Let me know if you want share some thoughts....

matthew-brett:
Thanks for your thoughts.  I don't know, but I'm guessing that, when a Python user wants to use symbolic computation, they usually turn to Sympy.  I guess you would also like them to consider Diofant.    So I suspect there are a large number of potential users, like me, who will want to know why they should use and contribute to Diofant instead of Sympy.  Obviously there is your difference in governance, but I think that only a few people realize how significant that is, and they tend to be very experienced users.   So the next question is - what can Diofant do that Sympy cannot?  And - is Diofant code easier to work on than Sympy code?   Is it going in a better direction - and if so - in what way?

The thing that made me think about this, just now, was the apparently near-intractable problem of the old and new assumptions in Sympy - see https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/sympy/8mwNjt4toEs/TMlwb36tBAAJ . How does Diofant comare?  But that's just an example...

skirpichev:
>    Thanks for your thoughts. I don't know, but I'm guessing that, when a
>    Python user wants to use symbolic computation, they usually turn to Sympy.
>    I guess you would also like them to consider Diofant. So I suspect there
>    are a large number of potential users, like me, who will want to know why
>    they should use and contribute to Diofant instead of Sympy.

[...]

PR #1133, opened 2021-03-19
Misc fixes

skirpichev:


skirpichev:
@mpeychev, your script from https://github.com/sympy/sympy/issues/21121 with obvious one-line modifications does work for Diofant:
```
$ cat a.py
import multiprocessing as mp
#import sympy as sp
import diofant as sp

VAR_X = sp.Symbol('x', real=True, nonzero=True)

def process():
    return sp.Symbol('x', real=True, nonzero=True)

if __name__ == '__main__':
    a1 = sp.Symbol('a', real=True, nonzero=True)
    a2 = sp.Symbol('a', real=True, nonzero=True)
    print(a1, a2, a1 == a2, a1 - a2, '\n')

    pool = mp.Pool(4)
    jobs = []
    for _ in range(5):
        jobs.append(pool.apply_async(process))
    symbols = []
    for job in jobs:
        symbols.append(job.get())
    pool.close()

    for s in symbols:
        print(s, ' | ', VAR_X, ' | ', s - VAR_X, ' | ', sp.simplify(s - VAR_X))
        print(hash(s), hash(VAR_X))
$ python a.py
a a True 0 

x  |  x  |  0  |  0
1101788483439232778 1101788483439232778
x  |  x  |  0  |  0
1101788483439232778 1101788483439232778
x  |  x  |  0  |  0
1101788483439232778 1101788483439232778
x  |  x  |  0  |  0
1101788483439232778 1101788483439232778
x  |  x  |  0  |  0
1101788483439232778 1101788483439232778
```

Maybe you can provide an example of the SymPy unit-test, that does fail?

mpeychev:
Thanks.

You mean to provide a minimal unit-test that tests this behaviour? I can do that.

Or to look at SymPy's codebase and find a relevant unit-test which is failing?

skirpichev:
> You mean to provide a minimal unit-test that tests this behaviour?

Yes, for SymPy.  Something, that may be included (as XFAILed) test in it's test suite.

mpeychev:
```
import sympy as sp

def get_symbol():
    return sp.Symbol('x', real=True)

def test():
    from multiprocessing import Pool

    x = sp.Symbol('x', real=True)
    with Pool(processes=1) as pool:
        job = pool.apply_async(get_symbol)
        s = job.get()

    assert (hash(s) == hash(x))
    assert (sp.simplify(s - x) == 0)


if __name__ == '__main__':
    test()
```

Here is a draft proposal. Unfortunately, it is not trivial to move the `get_symbol` method inside the test because the multiprocessing module then complains with: `AttributeError: Can't pickle local object...`

skirpichev:
> Here is a minimal unit test which reveals the issue in SymPy.

Thank you.  I think, there are something wrong in the Diofant too.  For your test:
```diff
diff --git a/diofant/tests/utilities/test_pickling.py b/diofant/tests/utilities/test_pickling.py
index c36b6774de..48dd4622e8 100644
--- a/diofant/tests/utilities/test_pickling.py
+++ b/diofant/tests/utilities/test_pickling.py
@@ -470,3 +470,15 @@ def test_sympyissue_7457():
     b = pickle.loads(pickle.dumps(a))
     b.evalf(strict=False)  # not raises
     assert a == b
+
+
+def test_sympyissue_21121():
+    x1 = Symbol('x', real=True)
+    x2 = Symbol('x', real=True)
+
+    assert hash(x1) == hash(x2)
+
+    x2_pickled = pickle.dumps(x2)
+    x2_unpickled = pickle.loads(x2_pickled)
+
+    assert hash(x1) == hash(x2_unpickled)
```
It pass if selected (`-k` pytest's option), but fails without.  Resetting cached `_hash` value helps: `x2_unpickled._hash = None`.  Maybe here should be used the standard caching mechanisms (`core.cache`), not yet another custom one...

skirpichev:
This should work as a quick fix:
```diff
$ git diff master diofant/core/
diff --git a/diofant/core/symbol.py b/diofant/core/symbol.py
index 93c07f0c03..8479a62b6a 100644
--- a/diofant/core/symbol.py
+++ b/diofant/core/symbol.py
@@ -141,7 +141,7 @@ def __getnewargs__(self):
         return self.name,
 
     def __getstate__(self):
-        return {'_assumptions': self._assumptions}
+        return {'_assumptions': self._assumptions, '_hash': None}
 
     def _hashable_content(self):
         # Note: user-specified assumptions not hashed, just derived ones
```

But, perhaps, it would be wise to reset this attribute somewhere in parent classes (Basic?).  @oscarbenjamin, does this works for SymPy?

oscarbenjamin:
> does this works for SymPy? (`_hash` is former `_mhash` attribute in SymPy)

Yes, it seems to. I'm not sure I've understood why that works though...

oscarbenjamin:
> > does this works for SymPy? (`_hash` is former `_mhash` attribute in SymPy)
> 
> Yes, it seems to. I'm not sure I've understood why that works though...

Oh, I see. The hash is somewhere computed (and cached) in between `Symbol.__new__` and `x._assumptions = ...` which is called implicitly by `loads`.

oscarbenjamin:
Probably using `__getnewargs_ex__` rather than `__getstate__` is a better fix.

skirpichev:
> Probably using __getnewargs_ex__ rather than __getstate__ is a better fix.

Maybe.  But in this case, probably, for Symbol's (and co) - we must save first user-defined assumptions. Right now - only derived assumptions are saved.

(As a side issue, I think that handling kwargs for Basic-derived classes should be standardized somehow, like for args.  But, perhaps, this is not a right time for this: I hope that `Symbol('x', real=True)` syntax will gone as I mentioned [here](https://github.com/diofant/diofant/issues/1028#issuecomment-643960291), i.e. in favor of `assume(Reals.contains(x))`.)

oscarbenjamin:
> Maybe. But in this case, probably, for Symbol's (and co) - we must save first user-defined assumptions. Right now - only derived assumptions are saved.

Why is it important to save only the user defined assumptions?

I can see that some users might find this confusing:
```python
In [22]: Symbol('x') == Symbol('x', real=True)
Out[22]: False

In [23]: Symbol('x', positive=True) == Symbol('x', real=True, positive=True)
Out[23]: True
```
Few users actually want to create different symbols with the same name and different assumptions though. I've only ever seen confusion comparing vanilla `Symbol('x')` with some symbol that has some assumptions set.

skirpichev:
> Why is it important to save only the user defined assumptions?

This will work (I think), if the assumption derivation is an idempotent operation, i.e. same set from the user assumptions and from the assumptions, derived from user assumptions...  Which was not last time I tried this in SymPy.

oscarbenjamin:
> Which was not last time I tried this in SymPy.

It should be. That would be a bug otherwise.

In the case of Symbol the `_assumptions` entry is not automatically populated with Nones but any assumption that can be derived should be populated straight away so the mapping from user-specified assumptions to `assumptions0` (which excludes the `None` predicates) should be always well-defined although many to one:
```python
In [95]: x = Symbol('x', real=True)

In [96]: x._assumptions
Out[96]: 
{'real': True,
 'imaginary': False,
 'finite': True,
 'complex': True,
 'extended_real': True,
 'hermitian': True,
 'infinite': False,
 'commutative': True}

In [97]: x.is_prime

In [98]: x._assumptions
Out[98]: 
{'real': True,
 'imaginary': False,
 'finite': True,
 'complex': True,
 'extended_real': True,
 'hermitian': True,
 'infinite': False,
 'commutative': True,
 'prime': None,
 'extended_negative': None,
 'extended_positive': None,
 'negative': None,
 'positive': None}

In [99]: x.assumptions0
Out[99]: 
{'real': True,
 'imaginary': False,
 'finite': True,
 'complex': True,
 'extended_real': True,
 'hermitian': True,
 'infinite': False,
 'commutative': True}
```

skirpichev:
FYI, @oscarbenjamin, your [patch](https://github.com/sympy/sympy/issues/21121#issuecomment-803212488) doesn't work for SymPy, yet (run pickle tests).  But I see reasons to cleanup pickling support for Basic-derived classes.

PR #1097, opened 2020-11-24
Implement a specialized class for FF(p)[x]

skirpichev:
- [x] adapt after #1103
- [ ] use DUP?
- [ ] docs
- [ ] release notes
- [ ] timings!

// for #966

PR #1153, opened 2021-06-24
Misc fixes

skirpichev:


skirpichev:
@user234683, I think sympy/sympy#13029 is fixed for the Diofant.  I've added test for your [last comment](https://github.com/sympy/sympy/issues/13029#issuecomment-322329070): all variables are taken into account.  Not sure how to test performance issue.  But I've these benchmarks:
```
In [4]: %timeit sqf_part(c0**4*c1**7*c2**5*c3**3 + 1, c0, c1, c2, c3)
3.07 ms ± 240 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %timeit sqf_part(c0**2*c1**5*c2**6*c3**4 + c1**3*c2**6 + 1, c0, c1, c2, c3)
3.53 ms ± 8.68 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [6]: %timeit sqf_part(c0**8*c1**8*c2**9*c3**9 + c0**7*c1**9*c2**6*c3**2 + c0**5*c1**9*c2**5*c3**4 + 1, c0, c1, c2, c3)
29 ms ± 130 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [7]: %timeit sqf_part(c0**9*c1**8*c2**6*c3**9 + c0**3*c1*c2**8*c3**7 + c0**2*c1**9*c2**8*c3 + c0**2*c1**6*c3**7 + 1, c0, c1, c2, c3)
33.5 ms ± 2.11 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [8]: %timeit sqf_part(c0**9*c1**7*c2**9*c3**5 + c0**4*c1**8*c2**7 + c0**3*c2**5*c3**6 + c0*c1**4*c2**7*c3**9 + c0*c2*c3**6 + 1, c0, c1, c2, c3)
37.2 ms ± 286 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [9]: %timeit factor(c0**4*c1**7*c2**5*c3**3 + 1, c0, c1, c2, c3)
305 ms ± 75.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [10]: %timeit factor(c0**2*c1**5*c2**6*c3**4 + c1**3*c2**6 + 1, c0, c1, c2, c3)
301 ms ± 106 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [11]: %timeit factor(c0**8*c1**8*c2**9*c3**9 + c0**7*c1**9*c2**6*c3**2 + c0**5*c1**9*c2**5*c3**4 + 1, c0, c1, c2, c3)
681 ms ± 73.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [12]: %timeit factor(c0**9*c1**8*c2**6*c3**9 + c0**3*c1*c2**8*c3**7 + c0**2*c1**9*c2**8*c3 + c0**2*c1**6*c3**7 + 1, c0, c1, c2, c3)
701 ms ± 168 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [13]: %timeit factor(c0**9*c1**7*c2**9*c3**5 + c0**4*c1**8*c2**7 + c0**3*c2**5*c3**6 + c0*c1**4*c2**7*c3**9 + c0*c2*c3**6 + 1, c0, c1, c2, c3)
858 ms ± 133 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
for your examples in the issue description.

Let me know if you think the issue is still valid.

user234683:
It's been a long time since that issue so I can't say for sure, but `sqf_part` looks good based on those numbers. In the original issue, there was a 1,000x - 100,000x increase in time between the smallest polynomial and the largest one, but you have it down to about 10x, which appears to be fine since the absolute time is small. Though the time for factor seems to be deteriorated. It was 48 ms in the original issue for the largest polynomial, but now takes 850 ms.

skirpichev:
> but you have it down to about 10x

Actually, it's smaller if you compare tests with same max_degree and with increasing max_degree.  Though I'm not sure how stable your numbers are: your script does only a single run for sqf_part().

> Though the time for factor seems to be deteriorated. It was 48 ms in the original issue for the largest polynomial, but now takes 850 ms.

Probably, that's because the Diofant doesn't use yet a special implementation for univariate polynomials over Galois fields (of prime order), as SymPy does.  I think, that can be mitigated by diofant/diofant#1097.

PR #1232, opened 2022-04-08
series: support directional limits on the complex plane

skirpichev:


anutosh491:
Could you check if this works correctly ? Maybe this would fail 
```
>>> limit(atan(x), x, I*oo)   # expected pi/2
-pi/2
```

skirpichev:
On Fri, Apr 08, 2022 at 05:43:26AM -0700, Anutosh Bhat wrote:
>    Could you check if this works correctly ? Maybe this would fail
> 
>  >>> limit(atan(x), x, I*oo)   # expected pi/2
>  -pi/2

$ python -m diofant

In [1]: limit(atan(x), x, I*oo)   # expected pi/2
Out[1]: 
π
─
2

BTW, this does work on the master, isn't?


This PR is far from completion, however.  I'm still thinking
about API.  And only default bahaviour was implemented (i.e.
if there is no `dir` - it's assumed to be `sign(z0)`)

anutosh491:
I am using 0.12.0 for basic reference , where it fails , yeah I realize this pr would take more effort than what is done currently !

skirpichev:
Latest "stable" is 0.13.0.

skirpichev:
@anutosh491, the SymPy's series() method has `cdir` and `dir` options.  Do you understand why?

PR #1298, opened 2023-02-23
calculus: fix handling nested Order() terms in Limit.doit()

skirpichev:
- [ ] adapt gruntz.py instead?
- [x] restrict to Add arguments?


skirpichev:
depends on #1296

PR #1337, opened 2023-07-22
Misc fixes

skirpichev:


skirpichev:
@dianetc, sorry, I can't reply you in the issue thread.  So I'll answer here.

> `res(f,g)==-1`

That's not true.  Usual definition (see e.g. [here](https://mathworld.wolfram.com/Resultant.html)) requires it should be ``(0-(-1))*(0-(-1))*(0-(-1))`` as we have ``x=0`` root for ``f=x**3`` (with multiplicity 3) and ``x=-1`` root for ``g=x+1``.  The ``-1`` is a correct answer for ``g=x-1``.

Perhaps, there could be a reversed sign convention, but as far as I know - most textbooks and another CAS use above one.  E.g. an example in the Mathematica:
```
Wolfram Language 13.3.0 Engine for Linux x86 (64-bit)
Copyright 1988-2023 Wolfram Research, Inc.

In[1]:= Resultant[x^3, x + 1, x]                                                                                                                               

Out[1]= 1
```

dianetc:
@skirpichev 
I was going by the [definition here](https://en.wikipedia.org/wiki/Resultant) , so $`resultant(f,g) = det \begin{pmatrix}  0 & 1 & 0 & 0 \\ 0 & 1 & 1 & 0 \\ 0 & 0 & 1 & 1 \\ 1 & 0 & 0 & 1\end{pmatrix} = -1`$ . In any case, [sagemath agrees with your computation](https://github.com/sympy/sympy/issues/25406#issuecomment-1646551031), so my perception of the ordering is wrong.

skirpichev:
> I was going by the [definition here](https://en.wikipedia.org/wiki/Resultant)

This article uses same definition, as in my reference above.

Your matrix is wrong.  Please note, that the leading coefficient is numbered by 0 in this article (i.e., ``f=[1, 0, 0, 0]`` and ``g=[1, 1]`` in a dense representation where the leading coefficient is first).  The correct matrix is:
```
In [1]: Matrix([[1, 1, 0, 0],
   ...:         [0, 1, 1, 0],
   ...:         [0, 0, 1, 1],
   ...:         [0, 0, 0, 1]]).det()
Out[1]: 1
```

PR #1347, opened 2023-08-25
Misc fixes

skirpichev:


skirpichev:
@sylee957, regarding sympy/sympy#25595 - there is a lot of other places, where ordered was used, e.g. TR9().  This was first introduced in b5e324b458, perhaps to reduce some indeterminacy in the algorithm.  I would appreciate if @smichr could the reason(s?).

Meanwhile, another approach could be masking all non-commutative parts of Mul expressions before using fu().

skirpichev:
BTW, removing all (well, except for two) ordered() calls in the fu.py doesn't break tests (https://github.com/diofant/diofant/pull/1348).  I think this is valid for the SymPy too.

PR #1415, opened 2024-06-11
polys: add RootOf.as_real_imag()

skirpichev:


diofant/polys/rootoftools.py
```diff
@@ -218,6 +218,11 @@ def _eval_conjugate(self):
             ci = self.index + 2*((self.index - nreals + 1) % 2) - 1
             return self._new(self.poly, ci)
 
+    def as_real_imag(self, deep=True, **hints):
+        if (cself := self._eval_conjugate()) is not None:
+            return ((self + cself)/2, (self - cself)/I/2)
+        return super().as_real_imag(deep=deep, **hints)
```
skirpichev, line 224: @oscarbenjamin, this approach should be less costly for SymPy too, c.f. https://github.com/sympy/sympy/pull/26828 (but to work correctly, https://github.com/sympy/sympy/issues/14293 should be solved in SymPy, not just closed).

BTW, which application do you see for Poly.real_imag() in multivariate case?

diofant/tests/polys/test_rootoftools.py
```diff
@@ -239,6 +239,13 @@ def test_RootOf_conjugate():
     assert r4.evalf() == -r5.evalf()
 
 
+def test_RootOf_as_real_imag():
+    r1, r2 = RootOf(x**3 + x + 3, 1), RootOf(x**3 + x + 3, 2)
+    assert r1.as_real_imag() == (r1/2 + r2/2, -I*(-r2 + r1)/2)
```
oscarbenjamin, line 244: If you evaluate these expressions numerically do you end up with a small imaginary part?

I think it is better if `expr.as_real_imag()[0].n()` evaluates as a definite real number so I would prefer `re(r1)` here.

skirpichev: > If you evaluate these expressions numerically do you end up with a small imaginary part?

Yes, I think so.

> I would prefer re(r1) here.

Hmm, it's possible with ``re(r1 + r2, evaluate=False).evalf()``.  But I doubt this is a good idea to return unevaluated expressions from as_real_imag().

Rather, evalf() should drop imaginary (or real) component if an expression is a real number (or pure imaginary).  Such test should work:
```diff
diff --git a/diofant/core/evalf.py b/diofant/core/evalf.py
index b3e6d39d3b..1747805b19 100644
--- a/diofant/core/evalf.py
+++ b/diofant/core/evalf.py
@@ -1214,7 +1214,7 @@ def evalf(self, dps=15, subs=None, maxn=110, chop=False, strict=True, quad=None)
             re = Float._new(re, p)
         else:
             re = Integer(0)
-        if im:
+        if im and self.conjugate() != self:
             p = max(min(prec, im_acc), 1)
             im = Float._new(im, p)
             return re + im*I
```

PR #1492, opened 2026-06-01
Misc fixes

skirpichev:


skirpichev:
CC @oscarbenjamin

I think that https://github.com/sympy/sympy/issues/23828 is a bug, not a documentation issue.  In the Diofant that was fixed long time ago with 9f3a6af3c3.  IIUIC, SymPy still uses DMP representation in some places and it seems that division algorithms are different for DMP and sparse polynomials.

The dense version looks wrong (or meaningless) for me.  In sparse version you should have same answers with *one* divisor (special case of reduction by the Groebner basis):
```pycon
>>> from sympy import ZZ, ring
>>> R, a,b,v,p,x = ring('a,b,v,p,x', ZZ)
>>> (a*x + b + p).div([a])
([x], b + p)
>>> R.dmp_rr_div(a*x + b + p, a)
(x, b + p)
>>> R, b,v,p,x,a = ring('b,v,p,x,a', ZZ)
>>> (a*x + b + p).div([a])
([x], b + p)
>>> R.dmp_rr_div(a*x + b + p, a)
(0, b + p + x*a)
```

